PR #11, opened 2018-09-26
Update dylib path on macos.

Carreau:
Dylib seem to have a leading `lib`, this make the repl, and kernel works on MacOS (at least if you manage to get ZMQ to compile)


Carreau:
I signed it!)

davidlattimore:
Thanks :)

PR #24, opened 2018-10-06
enhance jupyter packaging

davidB:
with logo, version in the banner and support of JUPYTER_PATH


davidlattimore:
Thanks for doing this!

Do you think it might be possible to move the images into a "third_party" directory? Any contributions where the contributor doesn't own the copyright need to go into a "third_party" directory. We should probably also include in the directory containing the images, information about where they came from, their license etc. Unfortunately we need to do this even if the license is Apache-2. Sorry for the hassle.

davidB:
Sorry, I'm not familiar with [Third-Party – opensource.google.com](https://opensource.google.com/docs/thirdparty/), I hope my last commit match requirements.

davidlattimore:
Thanks for your contribution. Like to add yourself to the authors list?

davidB:
>  Like to add yourself to the authors list?

Yes, Thank you.

PR #29, opened 2018-10-14
[WIP] Add helpers into evcxr_runtime

davidB:
Hi,

This PR is a try to extend runtime with helpers and shortcuts but I need your feedback about:
1. Do you want to have some helpers / ready to function like [gophernotes/Display.ipynb at master · gopherdata/gophernotes](https://github.com/gopherdata/gophernotes/blob/master/examples/Display.ipynb) as part of the lib. Is evcxr_runtime the right place to add helpers / shortcut to display content from outside (serde_json, images, vega, ndarrays,...) ? or do you prefer
    1. to have them into each own sub-project (enabling rust features is not accessible from jupyter kernel)
    1. to have them into an other project / repository (not under evcxr repository)
    1. other suggestions ?
1. If ok for helpers, to you have some guidelines you would like we follow ?
1. how to play in jupyter/repl with an unpublished crates (wip,...) ?

Tell me before I push other helpers function.

If you like I can move the cherry-pick the first commit into an other PR.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -33,6 +33,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "base64"
+version = "0.9.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "byteorder 1.2.6 (registry+https://github.com/rust-lang/crates.io-index)",
+ "safemem 0.3.0 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
 name = "bitflags"
 version = "0.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -54,6 +63,11 @@ dependencies = [
 [[package]]
 name = "byte-tools"
 version = "0.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
+name = "byteorder"
+version = "1.2.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
@@ -184,6 +198,9 @@ dependencies = [
 [[package]]
 name = "evcxr_runtime"
 version = "1.0.0"
+dependencies = [
+ "base64 0.9.3 (registry+https://github.com/rust-lang/crates.io-index)",
+]
 
 [[package]]
 name = "failure"
@@ -428,6 +445,11 @@ dependencies = [
 ]
 
 [[package]]
+name = "safemem"
+version = "0.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
 name = "sha2"
 version = "0.7.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


davidlattimore:
Thanks for working on this! My main concern with adding everything to the evcxr_runtime crate would be compile times. I'd worry that both the cold compile time and the warm compile times might increase with each additional dependency, so it'd be good if users of these helpers only had to pay the compile-time cost for what they actually used. Adding a dependency on a crate like mime is probably fine, assuming it's as light weight as I suspect it is. You might try using the ":timings" command to turn on printing of how long things take, then you can get a feel as to whether things slow down with an extra dependency.

As to where separate crates for supporting display of different types live, I'm not too worried. I'm OK with including them in the evcxr repository, in the repository of types that are being displayed, or in a completely separate repository.

A crate that handles the display of some type that is defined by a different crate wouldn't be able to implement evcxr_runtime::Display, due to Rust's orphan rules. It should however be able define it's own Display trait with the same evcxr_display method on it, then implement that trait for whatever types it likes.

The evcxr_runtime::Display trait is mostly only useful if a crate wants to implement display for its own types (likely with a cargo feature flag).

As for experimenting with changes locally, I think this should be possible by using
\:dep mycrate = { path = "/absolute/path/to/crate" }
then extern crate mycrate.

This has only been possible since [this commit](https://github.com/google/evcxr/commit/b80fc8a71c54f61f48cf7bfd0c4f0a7251866d75). Hopefully it works OK. Let me know if it doesn't.

davidB:
Thanks for response, I'll try `:dep`and `:timings` with my wip evcxr_runtime

davidB:
right, but without I could not display timings & image (or other content-type).
It should also allow to display several output from a cell.
Do you have a suggestions to identify when it's several format of the same content ?

davidB:
Maybe I try to change the signature of `pub fn execute(&mut self, to_run: &str) -> Result<EvalOutputs, Error>`to return a `Result<Vec<EvalOutputs>, Error>` ?

davidB:
Current wip.
[Imgur](https://i.imgur.com/vHsxS3u.png)

davidB:
tbh I don't know what is the right way. I like the current "only be one output, with possibly several different representations". But currently I don't see the case where it is provided, and I guess the different representations are computed by evcxr and could be not provided by the user, isn't it ?
Also I guess the case of timing could become not alone.
(time to sleep in my TZ, maybe I'll got an idea tomorrow morning or a solution in this thread ;-) )

davidlattimore:
Sleep well.

Perhaps we should change \:timing so that it writes to stdout instead of returning an output. Currently anything that the subprocess writes to stdout is sent to Jupyter. That's a separate channel than the outputs.

The slightly tricky part is that it's currently the parent process that does the timing. If we moved the timing to the child process, then made it print, I think that might work better. This would involve the parent sending a command to the child process to turn timing on/off.

davidB:
WDYT about not trying to merge timing into content and let jupyter/repl display it as they like ?

```rust
pub struct EvalOutputs {
    pub content_by_mime_type: HashMap<String, String>,
    pub timing: Option<Duration>,
}
```

davidlattimore:
Splitting stuff sounds sensible. Adding a timing field to EvcxrOutputs sounds good to me.

davidB:
Content of this PR was dispatched between
- other PR
- the new runtimes folder
- and I started a new project  [davidB/evcxr_displayers: A library of displayers (and helpers) for evcxr_jupyter.](https://github.com/davidB/evcxr_displayers) to experiment

davidB:
Can I used the name evcxr_displayers ? (I didn't publish to crates.io yet, I'll wait your approval)

davidlattimore:
That sounds reasonable to me

PR #38, opened 2018-12-24
Upgrade dependencies

davidB:
Maybe connecting a tool like dependabot could help keep / to be alerted when dependencies publish new version.
I never used dependabot with cargo workspace project, but it works well with my rust project.

I didn't try to upgrade syn & proc-macro2


davidlattimore:
Unfortunately it looks like this doesn't work on the current stable version of rust. This is probably due to https://github.com/rust-lang/rust/issues/45601 which was only recently fixed. I had lots of trouble with this bug during development and it's the reason why I've pinned most dependencies to exact versions. It also meant that I had to avoid using certain libraries like a proper toml parser, since they used custom derive. Once stable is updated, we'll need to think about how soon we're happy to drop support for older versions of rustc.

PR #40, opened 2018-12-28
allow multiple command and mixed command & code per cell in jupyter

davidB:
following discussion in https://github.com/google/evcxr/pull/39, this PR

1. allow to run multiple command per cells
    Imgur
2. allow to run mixed command and code
    Imgur
3. stop execution after the first error
    Imgur

The change is located into evcxr_jupyter module only, because the repl doesn't need it (it executes line by line).

I guess improvement (behavior or impl) could be done.


davidB:
One possible improvement could be to sent a batch of commands & code, let the executor choose the strategy (one by one,  process first every commands then code (and fallback to to empty code), ???).

davidlattimore:
Thanks

PR #42, opened 2019-03-01
fix repl on windows

dgriffen:
This fixes the repl on windows. The two missing pieces were that the cfg used the wrong name (`windows` not `window`). And the path also needed to include the rust sysroot in order to load the std dll.


dgriffen:
I signed it!

PR #43, opened 2019-03-02
fix eval statements with exports

dgriffen:
This fixes eval statements like `pub struct Foo;` on Windows.

To explain the fix: on Windows once you load a dll you are not allowed to overwrite the file. The previous build process would do just that if you wrote a statement with an export and then attempted to compile any further lines. The reason for this is that rustc will recompile a crate if it is used as a dependency vs just compiled on its own. To mitigate this a "meta crate" is created that lists all crates as its dependencies, this means that all of the crates will be compiled in the proper manner and rustc will not attempt to recompile them. This avoids the file locking issues on windows.

I have also tested this against linux (via WSL) and verified that it still works.

### DIFF
--- a/evcxr/src/eval_context.rs
+++ b/evcxr/src/eval_context.rs
@@ -44,6 +44,7 @@ pub struct EvalContext {
     pub(crate) debug_mode: bool,
     opt_level: String,
     next_module: Arc<Mutex<Option<Module>>>,
+    meta_module: Option<Module>,
     state: ContextState,
     child_process: ChildProcess,
     // Whether we'll pre-warm each compiled crate by compiling the same code as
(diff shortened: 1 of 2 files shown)


davidlattimore:
Thanks for this. Do the tests now pass on Windows? If they do, then trying to set up CI for Windows could be sensible to prevent regressions.

dgriffen:
running `cargo test` in the evcxr crate makes everything pass except for `abort_and_restart` which fails because windows gives a different exit code than linux. I can either make a PR to make a second identical test that is configured for windows or just disable the test entirely on windows. Your choice.

davidlattimore:
Good news. Apparently this change fixed stuff on macos as well :)

Regarding the test, a second version of the test for windows sounds best, but I'd be happy with either

PR #50, opened 2019-07-22
Add Cargo.lock

thomasjm:
I'm interested in packaging this tool using Nix. There was a long thread [here](https://github.com/NixOS/nixpkgs/issues/29981) about how to do so for a crate that doesn't include `Cargo.lock` in version control, and the conclusion seems to be that the easiest thing is to just add it :)

Since this project produces binaries, it seems to be recommended by the [Cargo book](https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries) to include it as well.


davidlattimore:
I did used to have a Cargo.lock, but deleted it in ddf609651ab578a35db5da01fbaae18a186d6694. My motivation at the time for deleting it was that Evcxr had become uninstallable due to a change in some upstream crates, but Travis didn't report a build failure because it was using the Cargo.lock. I also realised that the Cargo.lock, being at the workspace level wasn't actually being published to the individual crates on crates.io, so wasn't actually serving any useful purpose. If you know of a way to include Cargo.lock in the published crates, then I'd be all for that.

thomasjm:
Well, I don't actually know anything about Rust, but a few thoughts:

* Looking at the chapter on [Cargo Workspaces](https://doc.rust-lang.org/book/ch14-03-cargo-workspaces.html), it seems having a single `Cargo.lock` in the root workspace directory is the way it's supposed to work.
* Having the build break because of changes in upstream crates seems like precisely the reason you would want a lockfile :P. At least then you can clone the repo and always build successfully.
* Have you tried the `publish-lockfile` option described [here](https://doc.rust-lang.org/cargo/reference/unstable.html)?  The book says the following:

> When creating a `.crate` file for distribution, Cargo has historically not included the `Cargo.lock` file. This can cause problems with using `cargo install` with a binary. You can specify that your package should include the `Cargo.lock` file when using `cargo package` or `cargo publish` by specifying the `publish-lockfile` key in `Cargo.toml`.

Reading between the lines, I guess the traditional way to avoid build breakages is to tighten up the version bounds in `Cargo.toml`. But hopefully this `publish-lockfile` option works in the workspace scenario.

davidlattimore:
Funnily enough it was actually having tight bounds on versions in Cargo.toml that caused the previous breakage. I had locked down all of evcxr's direct dependencies to specific versions in order to avoid a rustc bug that was way too easy for upstream crates to trigger. Fortunately that rustc bug is now long fixed. But an indirect dependency made a change that then didn't work with the locked down version of my direct dependency. Pretty sure someone wasn't following semver correctly, but I didn't look into it too deeply, since it was easily fixed by just upgrading to the latest versions of some package.

Thanks for pointing out publish-lockfile. Unfortunately it looks like it's not available on stable Rust yet. Looks like it will be in next release though - 1.37. In fact it sounds like they've changed it to publish the lockfile by default for binary crates.

Where is it that you actually need Cargo.lock, is it in the crates published to crates.io or in the github repo?

thomasjm:
I need it in the Github repo--the Nix Rust build infrastructure gets confused if it doesn't find it.

davidlattimore:
OK. I'm happy to merge this then. I guess if I want to avoid having Travis build using it, I can probably add a step to the travis file to delete the file before building.

I can't unfortunately merge this without the CLA having been accepted. I know it's silly, especially for such a small change. If you'd rather not sign it, let me know and I'll close this PR and make the change myself.

thomasjm:
Thanks! I thought I had signed it in the past. Just did it again, hopefully it picks it up now...

davidlattimore:
Thanks. Good luck with the packaging :)

PR #68, opened 2019-11-18
Add flag for disabling readline

dima74:
Additionally fixes prompt color (if I understand correctly, it is supposed to be yellow)


davidlattimore:
Thanks!

Issue #69, opened 2019-11-18
I keep getting libctx error every expression


GrayJack:
When I start evcxr i get this error
`Error renaming '"/tmp/.tmpv1xw2x/target/debug/deps/libctx.so"' to '"/tmp/.tmpv1xw2x/target/debug/deps/libcode_1.so"': No such file or directory (os error 2)`

But it also happens after every expression put in the repl

davidlattimore:
Hi @GrayJack . Thanks for the report. To help figure out what's happening, would you be able to do the following?

- From evcxr, after entering an expression, type `:last_compile_dir`
- In a separate terminal, cd to the directory that it printed.
- Check to see if target/debug/deps/libctx.so exists.
- What about target/debug/libctx.so ?
- Run `cargo rustc --message-format=json -- -C prefer-dynamic`
- After that, which of those files exists?

Also possibly useful:

- Output of rustc --version
- What OS / distribution are you running?

On my system both files are created (target/debug/deps/libctx.so and target/debug/libctx.so). Actually, they're the same file, but one is hard linked to the other. I'm not sure why I chose to make use of the one in the "deps" directory. I note that cargo now emits some JSON that says what file was created (for me, it indicates the non-deps directory filename). Perhaps I should switch to reading that JSON.

GrayJack:
Sorry, forgot about those infos
Rustc: rustc 1.40.0-nightly (1423bec54 2019-11-05)
OS: Linux, version 5.3.10

> From evcxr, after entering an expression, type `:last_compile_dir`
It gives me the same error
```sh
Error renaming '"/tmp/.tmpL77Dg7/target/debug/deps/libctx.so"' to '"/tmp/.tmpL77Dg7/target/debug/deps/libcode_3.so"': No such file or directory (os error 2)
```

> In a separate terminal, cd to the directory that it printed.
The max it can go is `/tmp/.tmp<random>/` with the following tree
```sh
.
├── Cargo.lock
├── Cargo.toml
└── src
    └── lib.rs
```

> Check to see if target/debug/deps/libctx.so exists.
>  What about target/debug/libctx.so

Because of the above, neither exist

Maybe it's related to me having a custom target folder for my rust environment?

> Run `cargo rustc --message-format=json -- -C prefer-dynamic`

Yeah, that proves it, it is because of my custom target folder
```
[...]

PR #74, opened 2019-12-01
Specify target-dir via environment variable CARGO_TARGET_DIR

Aloxaf:
I came across the same problem as https://github.com/google/evcxr/issues/69, but https://github.com/google/evcxr/commit/dd41f471bc8953681137ec2596df86bfcf53efab didn't fix it for me. Because I specified the target-dir via `~/.cargo/config`.

I think the best way to solve this problem is to let evcxr specify the `target-dir` by itself, via `CARGO_TARGET_DIR`(it has a higher priority)

### DIFF
--- a/evcxr/src/module.rs
+++ b/evcxr/src/module.rs
@@ -167,7 +167,7 @@ impl Module {
             .arg("--")
             .arg("-C")
             .arg("prefer-dynamic")
-            .env_remove("CARGO_TARGET_DIR")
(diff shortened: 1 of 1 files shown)


Aloxaf:
@googlebot I signed it!

googlebot <notifications@github.com> 于2019年12月1日周日 下午11:13写道：

davidlattimore:
Thanks!

Issue #67, opened 2019-11-11
Allow customizing format output


thomcc:
I often want to sanity check some math or similar, and evcxr_repl works wonderfully for this for the most part (previously I had been using lldb/gdb for this which works about as well as you might imagine...).

Unfortunately, if doing bitwise math, I often want to see hex output. Doing so requires outputting something like `format!("{:x}", <expression I want to check>);` each time, which is tedious.

Taking a look at the source, it seems somewhat tricky since you don't know the type in advance, but one (somewhat hacky) option would be something like allowing specifying a list of formatting options to try, which would be used [here](https://github.com/google/evcxr/blob/e410ecb969e50cdffae7d837a59813cade232b0c/evcxr/src/eval_context.rs#L225), or something. e.g. if I set `:fmt ["{:x}", "{:?}"]` (or whatever), then the system would try {:x} followed by {:?} for things where evcxr_display doesn't apply.

That said, I'm flexible and willing to accept something that requires annotating things manually if the above solution is undesirable for some reason.

davidlattimore:
Interesting. I'd not rule out something like that if there was sufficient demand for it. In the meantime, perhaps the following would suffice?

`pub trait HexDisplay { fn evcxr_display(&self); } impl HexDisplay for i32 { fn evcxr_display(&self) {println!("EVCXR_BEGIN_CONTENT text/plain\n0x{:x}\nEVCXR_END_CONTENT", self);} }`

This adds a custom text display format for i32, after which i32 values will be printed as hex:
```
>> 8191
0x1fff
```

thomcc:
For anybody interested, I actually ended up solving this for myself with a "prelude" of sorts in my init.evcxr: https://gist.github.com/thomcc/8b308c30ddc35eb07646b4bb2ade480f (contains examples)

The downside of this is that init.evcxr is kinda a pain to work with (long lines), but other than that it works like a charm.

davidlattimore:
Wow that's a lot of code to put on one line. Fixing it might just be a matter of removing the line splitting code from the init.evcxr handling code. I think at the the time that code was written evcxr didn't support an input containing multiple commands, but now it does.

PR #77, opened 2019-12-07
let output format customizable

KeenS:
It works like this.  CC #67.

```
Welcome to evcxr. For help, type :help
>> :help
:vars             List bound variables and their types
:opt [level]      Toggle/set optimization level
:fmt [format]     Set output formatter (default: {:?}).
:explain          Print explanation of last error
:clear            Clear all state, keeping compilation cache
:dep              Add dependency. e.g. :dep regex = "1.0"
:sccache [0|1]    Set whether to use sccache.
:version          Print Evcxr version
:preserve_vars_on_panic [0|1]  Try to keep vars on panic

Mostly for development / debugging purposes:
:last_compile_dir Print the directory in which we last compiled
:timing           Toggle printing of how long evaluations take
:last_error_json  Print the last compilation error as JSON (for debugging)
:time_passes      Toggle printing of rustc pass times (requires nightly)
:internal_debug   Toggle various internal debugging code

>> #[derive(Debug, Clone, Default)] struct Foo { bar: i32 }
>> Foo::default()
Foo { bar: 0 }
>> :fmt {:#?}
Output foramt: {:#?}

>> Foo::default()
Foo {
    bar: 0,
}
>> :fmt {}
Output foramt: {}

>> Foo::default()
   ^^^^^^^^^^^^^^ `Foo` cannot be formatted with the default formatter
`Foo` doesn't implement `std::fmt::Display`
help: the trait `std::fmt::Display` is not implemented for `Foo`
>> 1
1
>> "hogehoge"
hogehoge
>> :fmt 0x{:x}
Output foramt: 0x{:x}

>> 32
0x20
>>
```

### DIFF
--- a/evcxr/src/command_context.rs
+++ b/evcxr/src/command_context.rs
@@ -158,7 +158,7 @@ impl CommandContext {
                 let new_format = if let Some(f) = args { f } else { "{:?}" };
                 self.eval_context.set_output_format(new_format.to_owned());
                 text_output(format!(
-                    "Output foramt: {}",
+                    "Output format: {}",
                     self.eval_context.output_format()
                 ))
             }
--- a/evcxr/src/command_context.rs
+++ b/evcxr/src/command_context.rs
@@ -154,6 +154,14 @@ impl CommandContext {
                 self.eval_context.set_opt_level(new_level)?;
                 text_output(format!("Optimization: {}", self.eval_context.opt_level()))
             }
+            ":fmt" => {
+                let new_format = if let Some(f) = args { f } else { "{:?}" };
+                self.eval_context.set_output_format(new_format.to_owned());
+                text_output(format!(
+                    "Output foramt: {}",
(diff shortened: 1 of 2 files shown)


KeenS:
@googlebot I signed it!

thomcc:
I'm not sure how useful it is to do this in this format. In #67 I suggested there be a list of fallbacks, so that setting the format to {:x} wouldn't mean that only things that impl std::fmt::LowerHex could be used... Unless I'm mistaken that's a drawback of the current setup.

That said, I guess as-is it could still be useful for e.g. setting the default to {:#?} or something

davidlattimore:
Thanks. Being able to use {:#?} seems like a good use-case for this change. Happy to merge this. Will wait a bit to give you a change to fix the typo that @bjorn3 noticed. Also happy to just merge and then I can follow up with a fix for the typo.

Issue #78, opened 2019-12-07
Allow customizing config dir


thomcc:
On macOS this uses `~/Library/Preferences/evcxr`. I'd prefer this to be in my home dir (ideally `~/.config/evcxr` like on other unixes), as this is consistent with other CLI tools and etc. (`~/Library/Preferences` is really mostly used for GUI applications...).

Allowing it to be overridden by a `EVCXR_CONFIG_DIR` environment variable would be nice and not break anybody, I could file a PR for this if you'd accept it.

davidlattimore:
Evcxr just defers to the dirs crate - https://crates.io/crates/dirs - to determine where to read config files from. Perhaps it's ideas of where things should go on different platforms is out of step with those platforms? But, regardless, I'd be happy to accept a PR for adding an EVCXR_CONFIG_DIR.

thomcc:
> Perhaps it's ideas of where things should go on different platforms is out of step with those platforms

Yeah, not sure. Generally things use `Library/Application Support` (which it calls `data_dir()`), but even then I generally don't expect things that aren't `.app` bundles to write into `~/Library` (or `/Library`). Either way, it's not like they can really change it without breaking anybody who uses it currently (same situation as with std::env::home_dir...).

PR #81, opened 2019-12-07
Run prelude.rs after init.evcxr if one exists, and support specifying config path via environment.

### LINKED ISSUE #79: Support loading a prelude file?
As I mentioned in https://github.com/google/evcxr/issues/67#issuecomment-562889010, it's a bit annoying to 

Currently one option would be for me to define this in a separate crate (say 'thom_evcxr_prelude'), use `:dep thom_evcxr_prelude { path = "/Users/thom/src/thom_evcxr_prelude }` followed by `use thom_evcxr_prelude::*`.

However, when I tried this it impacted startup time of evcxr noticeably, and so I chose to just put everything on one line, since I start evcxr much more often than I touch that file. Defining a separate crate also feels like overkill if you have just, say, a small number of declarations and use statements which might be too involved to conveniently fit on one line (the initial suggesion in https://github.com/google/evcxr/issues/67#issuecomment-553833421 for example)

An easy solution would be just to have `<evcxr config dir>/prelude.rs` which is loaded as a single rust expression after init.evcxr runs (doing it after allows init.evcxr to include a dep, which the prelude uses, for example).

Alternatively, a `:load <path/to/blah.rs>` might be a more general useful feature, and would allow splitting to multiple files, and solves the 'when should it run' question, but seems like it might confuse users who try to point it at a lib.rs or mod.rs or something, and have it cause problems that way.

I would be happy to submit a PR for this if it would be accepted.

thomcc:
Tested manually (both with and without the prelude), since there aren't automated tests for this I could find.

### DIFF
--- a/evcxr/src/lib.rs
+++ b/evcxr/src/lib.rs
@@ -39,3 +39,14 @@ pub use crate::command_context::CommandContext;
 pub use crate::errors::{CompilationError, Error};
 pub use crate::eval_context::{EvalContext, EvalContextOutputs, EvalOutputs};
 pub use crate::runtime::runtime_hook;
+
+/// Return the directory that evcxr tools should use for their configuration.
+///
+/// By default this is the `evcxr` subdirectory of whatever `dirs::config_dir()`
+/// returns, but it can be overridden by the `EVCXR_CONFIG_DIR` environment
+/// variable.
+pub fn config_dir() -> Option<std::path::PathBuf> {
+    std::env::var_os("EVCXR_CONFIG_DIR")
+        .map(std::path::PathBuf::from)
+        .or_else(|| dirs::config_dir().map(|d| d.join("evcxr")))
+}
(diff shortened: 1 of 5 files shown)


thomcc:
Ah, I missed that you said you'd accept #78, one minute and I'll add a commit for that too (since otherwise it would conflict).

thomcc:
Note: in case it matters, I tested with evcxr_repl, I'm not sure how to set up the jupyter stuff to test evcxr_jupyter, but I have no reason to think it wouldn't work (but I am willing to try if you think it's needed).

evcxr/src/lib.rs
```diff
@@ -39,3 +39,14 @@ pub use crate::command_context::CommandContext;
 pub use crate::errors::{CompilationError, Error};
 pub use crate::eval_context::{EvalContext, EvalContextOutputs, EvalOutputs};
 pub use crate::runtime::runtime_hook;
+
```
thomcc: Wasn't sure what the best place for this function was. `lib.rs` is mostly just re-exports, but nothing else seemed that relevant, and it didn't seem worth adding a new module just for this function. Let me know if you'd like me to move or change this somehow though.

davidlattimore:
Thanks!

PR #82, opened 2019-12-12
Add flag for integrating Evcxr with Intellij Rust

dima74:
Hello! Currently I am working on [integrating Evcxr with Intellij Rust](https://github.com/intellij-rust/intellij-rust/pull/4615). Everything is great so far, and one of the remaining tasks is to determine status of last executed command (was it successfully compiled or not).

One of the possible solutions is to add "ide mode" to Evcxr. In this mode Evcxr will print special char before each prompt (`'\x01'` if compilation was successful and `'\x02'` otherwise).

I would say changes to the code are relatively small (except for the addition of `structopt` crate - if you want I can replace it with other library or some sort of manual parsing).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6,6 +6,14 @@ version = "1.0.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
+name = "ansi_term"
+version = "0.11.0"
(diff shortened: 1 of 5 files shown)


davidlattimore:
Cool. I definitely approve of the use of structopt. It's a very nice library. Good luck with the Intellij integration and feel free to reach out if you have any other requests or questions.

PR #98, opened 2020-01-13
Multiline support for evcxr_repl

### LINKED ISSUE #4: Add multiline support in the REPL
Can possibly use syn to parse what's been written and keep asking for more lines of input until syn reports that there's a complete token tree.

Alternatively, perhaps we can use a key modifier - e.g. shift+enter to execute. Perhaps need to first see what other REPLs do. Jupyter notebook uses shift+enter to execute, but then it's not a command line REPL (although jupyter-console is)

thomcc:
Fixes #4

This is something I've wanted for a while, and have been poking at off and on.

It ended up being a lot more involved than I had hoped. This is because essentially nothing that parses or scans rust seems to handle unmatched brackets at all, since they all operate on token trees, which must be well formed. I even asked wg-grammar channel on discord, and basically got told that I was going to have to go my own way, or use something like tree-sitter (which would only solve half the problem (lexing), requires taking a C dependency, and then either writing a tree-sitter grammar for rust or vendoring one from one of the text editors...

I considered the naive solution of just scanning for parens/brackets and ignoring anything else, but in practice it's pretty annoying: it's not really okay for the repl to see a `"("` and then never send your input since it thinks it's invalid -- so you need to at least lex the input well enough to notice strings/chars/comments, so that's what I did.

Unfortunately, even just this much is pretty involved, as rust's syntax for these is surprisingly tricky: Block comments can nest, chars are hard to distinguish from lifetimes, and it has `raw` strings.

That said, it works well in practice, has lots of comments, and pretty good test coverage, so hopefully it's not too controversial.

I'd also like to improve the repl more in the near future -- rustyline seems to have support for a few more things that would be useful, and it's possible we can use `syn` to handle many of the cases this misses (in particular detecting missing semicolons on stuff like `let x = 3` probably requires syn in order to be robust), but for now this patch was already pretty huge, so I left it as is.

Note: I've only tested this manually on macos at the moment, but have access to other platforms and will try to get around to it soon.

### DIFF
--- a/evcxr_repl/Cargo.toml
+++ b/evcxr_repl/Cargo.toml
@@ -13,9 +13,10 @@ edition = "2018"
 
 [dependencies]
 evcxr = { version = "=0.4.7", path = "../evcxr" }
-rustyline = "5.0.2"
+rustyline = "6.0.0"
 colored = "1.8.0"
 lazy_static = "1.3.0"
 failure = { version = "0.1.5", default-features = false, features = [ "std" ] }
 regex = { version = "1.3.1", default-features = false, features = [ "std" ] }
 structopt = "0.3"
+unicode-xid = "0.2"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ name = "ansi_term"
 version = "0.11.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "winapi 0.3.7 (registry+https://github.com/rust-lang/crates.io-index)",
+ "winapi 0.3.8 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
@@ -37,7 +37,7 @@ version = "0.2.13"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "libc 0.2.62 (registry+https://github.com/rust-lang/crates.io-index)",
- "winapi 0.3.7 (registry+https://github.com/rust-lang/crates.io-index)",
+ "winapi 0.3.8 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
@@ -75,7 +75,7 @@ dependencies = [
 
 [[package]]
 name = "bitflags"
-version = "1.1.0"
+version = "1.2.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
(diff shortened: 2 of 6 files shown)


evcxr_repl/Cargo.toml
```diff
@@ -13,9 +13,10 @@ edition = "2018"
 
 [dependencies]
 evcxr = { version = "=0.4.7", path = "../evcxr" }
-rustyline = "5.0.2"
+rustyline = "6.0.0"
```
davidlattimore: Looks like rustyline 6 uses the `non_exhaustive` feature, which was only stabilised in 1.40.0. I think I'm OK with dropping support for older versions of rust so long as it gives us something. Can you update .travis.yml to bump the minimum supported version to 1.40.0?

davidlattimore:
I'd like to merge this. Let me know if you'd like more time to address the comments I made before, otherwise I'll just merge and attempt to address them myself. Thanks again for this awesome change and for such well commented code :)

thomcc:
Sorry, meant to get back to this, not at my computer this weekend, so feel free to land it yourself!

cata0309:
How to use this multiline feature? The `\` is not working

thomcc:
It's automatically multiline if there are unclosed items, e.g. open parentheses, brackets, strings, etc.

thomcc:
Note that some issues with it exist, and are blocked on https://github.com/kkawakam/rustyline/pull/334. (As well as other bugs in that repo)

wangjiawen2013:
@thomcc @davidlattimore 
Hi,
Could you kindly enough to show how to support multi-line input in evcxr ?  Here is my evcxr interface and I can only input a line each time. What I want is to input a chunk of code consisting of multilines:

wangjiawen2013:
@thomcc @davidlattimore 
Hi,
Could you kindly enough to show how to support multi-line input in evcxr ?  Here is my evcxr interface and I can only input a line each time. What I want is to input a chunk of code consisting of multilines:

wangjiawen2013:
@thomcc @davidlattimore 
Hi,
Could you kindly enough to show how to support multi-line input in evcxr ?  Here is my evcxr interface and I can only input a line each time. What I want is to input a chunk of code consisting of multilines:

wangjiawen2013:
@thomcc @davidlattimore 
Hi,
Could you kindly enough to show how to support multi-line input in evcxr ?  Here is my evcxr interface and I can only input a line each time. What I want is to input a chunk of code consisting of multilines:

davidlattimore:
Hi @wangjiawen2013 

From the REPL, the only way currently to span multiple lines is for the previous like to be incomplete. e.g. if you have an open `(`, `{` etc. There's not currently any way to enter multiple lines without the previous lines being incomplete.

If you use the Jupyter kernel however, then you can enter multiple lines, so perhaps that might be an option.

PR #106, opened 2020-03-10
Update rustyline and highlighter logic

martell:
This will require a new release of rustyline but it would be a good point to start discuss what would make the repl more user friendly when doing multi line work.

Currently this

- changes the prompt to `.. ` when doing a multi line input
- highlights the previous bracket in blue after typing the closing match

This is a little rough around the edges and am not sure if `EvcxrRustylineHelper` is where we want to store the prompt configuration ?

### DIFF
--- a/evcxr_repl/src/repl.rs
+++ b/evcxr_repl/src/repl.rs
@@ -17,7 +17,7 @@ use colored::*;
 use rustyline::{
     completion::Completer,
     error::ReadlineError,
-    highlight::Highlighter,
+    highlight::{Highlighter, MatchingBracketHighlighter, PromptInfo},
     hint::Hinter,
     validate::{ValidationContext, ValidationResult, Validator},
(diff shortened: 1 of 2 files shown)


evcxr_repl/src/repl.rs
```diff
@@ -46,9 +54,33 @@ impl Highlighter for EvcxrRustylineHelper {
...
+            prompt.into()
+        }
+    }
+
+    fn has_continuation_prompt(&self) -> bool {
+        true
+    }
+
+    fn highlight_hint<'h>(&self, hint: &'h str) -> Cow<'h, str> {
+        hint.blue().to_string().into()
```
thomcc: We don't use `hinter` right now. I'd probably prefer waiting until we use it to decide it's styling (blue would be pretty jarring as a autosuggesion color anyway).

PR #110, opened 2020-03-18
Syntax highlighting (draft)

### LINKED ISSUE #107: Syntax highlighter support
This is still a bit off but I'd rather mention that I have this working locally so someone else doesn't end up wasting time on the same thing (#106 had this a little but was pretty mild).

I think the work on my machine might violate the license of the theme or syntax file I'm using... so I'm not going to upload a branch with this until I figure out if that's true, and if so resolve it. That said, you can see it being pretty here:

[asciicast](https://asciinema.org/a/3AbGHEf9OPvVsFnldNgWab9Nt)

It just uses syntect to do this, and was annoyingly tricky to actually make work well since that library is not geared at terminal programs, sadly.

thomcc:
WIP. Just based on the output of lexing (from rustc_lexer), with manually added support for keywords.

Still to do:

- [ ] Allow disabling syntax highlighting, possibly allow some configuration of it. (In practice it looks good on both light and dark screens, so maybe this can wait)
- [ ] Replace custom scanning code in scan.rs with rustc_lexer (not really required, but it's not great to have both).
- [ ] Comments for highlighting code -- it's pretty opaque probably.

That said there are a lot of rustyline bugs I noticed, so I'm probably going to push on that side a little before coming back to this... (also might be busy with work, but wanted to get this up sooner).

Fixes #107.


davidlattimore:
Nice :) Happy to merge once it's ready

aminroosta:
I hope this PR gets some more love ❤️ .
Thank you all 🙏

thomcc:
Yeah, I really need to get back to it...

thomcc:
I have a bit more time on my hands given the [recent news](https://techcrunch.com/2020/08/11/mozilla-lays-off-250/). One of the things I'm hoping to do for a bit is catch up on some of the open source stuff I have in a half-finished state, and this is probably the biggest bit that doesn't apply to a crate I maintain myself...

There were a few open questions here, mostly around configuration. It seems likely that users might want to disable this for various cases, for example: if it doesn't work with their terminal or whatever. This should be able to be a --flag, but I also think it should be configurable more permanently, and ideally would be configurable enough that I wouldnt have to hardcode the styles and escapes (which was enough for a prototype but wouldn't work well at large).

Something more dynamic would be better, and at a minimum we would want separate `light`/`dark` options (this can be autodetected on xterm-compatible terminals, but as much as people like to pretend otherwise... the world isn't made up of xterms)... In general, I don't really see a way out of this requiring a bit of external config.

Anyway, A natural way to configure things about `evcxr` is via `:foo` commands, since it can work at runtime, or be in the init.evcxr.

The issue here is: there isn't a natural way for the repl to accept these kinds of configuration (This would have been useful for #130 as well, as mentioned https://github.com/google/evcxr/issues/130#issuecomment-663609956). It *could* just intercept it before going to the main `evcxr` executable, but then it wouldn't be part of `:help`...

That said, even if it worked, it probably wouldn't be good for anything beyond a couple flags, and this doesn't feel like it really needs to be part of the runtime config, so I considered adding a `repl.toml` to the `evcxr` config dir for this, but...

That felt like too big of a change to just drop on you without discussing. I ultimately do still feel like some solution to allow repl-specifci `:foo` options will be needed eventually, but don't think that scales for complex things.

So yeah, needed to ask: for this kind of configuration, would you be alright with some sort of `repl.toml` for repl-specific configuration?

---

So, I feel like there's an elephant in the room, since you'll possibly notice that the demo in https://github.com/google/evcxr/issues/107#issue-578939617 uses custom theming and supports well known themes like Monokai, wheras this PR does not.

## Why this patch doesn't use `syntect`

Ultimately, I found reusing syntect (used by `bat` and friends) to be a rats nest full of complexity that kept piling on top of itself. Here are some of the issues:

- I never managed to get it to build on windows due to the native dep on oniguruma. I'm sure someone has made this work, but I couldn't make it happen. Some pointers indicate that maybe it's just a recent version that has an issue? IDK. Etiehr way, disabling this on windows is a possibility, but would be bad.

- Lots of terminals still don't support truecolor. Syntect styles produce give truecolor values, so you need to translate them. `bat` et al use a [LGPL library](https://crates.io/crates/ansi_colours) for this (and given that rust will statically link it in...), which seems non-viable for evcxr.

    That said, IMO that library has a bucnh of problems anyway, so I started working on my own: https://gitlab.com/thomcc/near-ansi (my gitlab is for unfinished projects)

    - Mine uses a more accurate color distance algorithm than theirs (CIE ΔE°76, which is by no means state-of-the-art any more, but offers a decent balance between performance and correctness. I didn't look too closely but it seemed like theirs was ad-hoc and not even handling sRGB correctly, let alone making any attempt at perceptual correctness).

    - Mine supports more terminals (including 88color ones, not that these are common anymore).

    - Mine is pure-rust, when theirs weirdly is in C that is compiled to Rust.

    - That said, mine needs a bit more polish, and felt weird to introduce when we're already using one library for terminal coloring... which... I think has a cute/nice API but don't have super high opinions about beyond beyond that. (At least it finally supports truecolor/256color in the new versions...).

    Even after this, cases like `bat` still have to hack in extensions to tmTheme to get something that's viable for terminals. I don't remember the details though...

- Syntect themes and syntax defs are slow enough to parse that what everybody does is serialize them into a binary blob using `bincode` and load from that.

    This is pretty much a pain in the ass since there's easy tool for producing these, and so stuff like `bat` has `bat cache --build` to do so. They also want to support including the blob via `include_bytes!`, and they have a weird bootstrapping process as a build step. I didn't really love the thought of introducing this, since... it sucks.

- I wasn't able to find out what license the various `tmTheme`s `bat` and friends use were. A lot are various CC share-alike, but... a lot just have no author information. This... probably means they're not actually okay to distribute D:

    I did consider looking in likely install locations for them on the users computer, and did find a way to get them out of `bat` installs if the user had run `bat cache --build`, and even considered detecting bat and running it for you. This... seems dumb, unreliable, and sketchy though, and not everybody will have bat installed.

    Certainly some of these are open source though, but the ones that are don't look great with rust / in a terminal. I've made lots of editor themes and don't mind writing a theme for this (obviously I chose colors in the draft patch thats up now) but ultimately if I'm writing my own theme, what was I getting out of syntect?

- ... I dont remember everything, but this isn't even the complete list!

But then I just realized that this was a crapload of work that was being done to just allow us to use syntect! And if I just stop trying to use it, the problem gets simpler.

So I did the simple thing, but by this point was kind of tired enough of the problem that I lost motivation to push it over the finish line. That said, Sorry for letting it sit without updates for so long.

davidlattimore:
I'm was very sorry to hear about the situation at Mozilla, and sorry to hear that it affected you too.

Regarding `:foo` commands. I'd be open to some refactoring to allow commands that are specific to the repl (or jupyter kernel). That, said, I'd also be OK with having a config file specifically for the repl if you think that's a good option. I'll defer to yourself and anyone else who cares to chime in.

Also fine with whatever highlighting solution you decide. I don't think there's too much use of the colored crate, so if you add an equivalent crate, we can hopefully switch the existing uses over without too much trouble. Having two such dependencies isn't the end of the world anyway (although a smaller dependency tree is always nice). I agree having an LGPL dependency would probably be a problem (I tried to argue for this project to be dual license, but unfortunately couldn't convince the legal people).

If you need any help from me, let me know and I'll see what I can do

bjorn3:
@davidlattimore Did you push a new main branch and delete the old master branch instead of using github's branch rename option? This PR, #106 and #152 got closed due to the target branch getting deleted.

davidlattimore:
That was indeed what I did and I hadn't realized it had done that. Thanks for noticing. Should be fixed now

thomcc:
Okay, that prod got me to work on this again. Not ready for review, but should be in the near future, I'll unmark as draft when it is.

Milo123459:
Hey, anyway I can help out with this? Maybe we could use rust-analyzers syntax highlighting capabilities?

thomcc:
A big part of the issue is that this runs into a bunch of problems with rustyline, since it doesn't handle the escape sequences very robustly. The issue isn't so much the syntax highlighting as that.

That said, I should be able to get back to this in a week or two to try a couple things out.

Milo123459:
> A big part of the issue is that this runs into a bunch of problems with rustyline, since it doesn't handle the escape sequences very robustly. The issue isn't so much the syntax highlighting as that.
> 
> That said, I should be able to get back to this in a week or two to try a couple things out.

Great! Can't wait.

thomcc:
I came across https://crates.io/crates/reedline recently, which is promising (as mentioned, one of the reasons I haven't done much here is rustyline bugs). After the holidays, I'll see if that allows me to make progress here.

maxall41:
Any updates on this?

PR #111, opened 2020-03-27
Enable virtual terminal on windows when using --disable-readline flag

dima74:
When using --disable-readline mode colors are not displayed properly on windows (https://github.com/intellij-rust/intellij-rust/issues/5058). As proposed [here](https://github.com/mackwic/colored/issues/59), it can be fixed by enabling virtual terminal

Also changed success/failure marker characters from \u0001 and \u0002 to \u0091 and \u0092, because current marker characters tend to not work on windows

(Both changes are necessary to fix IntelliJ Rust integration with Evcxr on windows)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2,1376 +2,1347 @@
 # It is not intended for manual editing.
 [[package]]
 name = "adler32"
-version = "1.0.3"
+version = "1.0.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


davidlattimore:
Sorry about the slow response and thanks for the PR

dima74:
@davidlattimore thank you! Could you please make new release if possible

davidlattimore:
Sure. 0.5.1 has been released

Issue #114, opened 2020-04-15
Corrupted error messages in latest v0.5.0 on Windows


MirkoCovizzi:
Version: 0.5.0
CPU: Ryzen 1700
OS: Windows 10
Toolchain: x86_64-pc-windows-msvc

How to recreate the error:

1. Install evcxr: `cargo install evcxr`
2. Wait for the REPL to load completely
3. Write something that is not valid Rust, for example `error`
4. Corrupted error messages will appear, with consistent behavior

Example:

```
C:\Users\Mirko\Desktop>evcxr
Welcome to evcxr. For help, type :help
>> error
   [91m^^^^^[0m [94mnot found in this scope[0m
[91mcannot find value `error` in this scope[0m
>>
```

davidlattimore:
Thanks for the report. Do you know if it worked OK on a previous version? I unfortunately don't have a way to test on Windows.

bjorn3:
This is likely https://github.com/mackwic/colored/issues/59. What you see is not corruption, but uninterpreted ansi escape codes, which are used on unix to for example give colors to the terminal output.

MirkoCovizzi:
Never used it before recently (installed thanks to the new JetBrains Rust REPL integration).

If someone can't point me in some direction I could try to find a way to fix it and make a PR.

MirkoCovizzi:
@bjorn3 

Yep, that's it. In a Python codebase I hardcoded this to achieve that effect (also on Windows)

Red:
FAIL = '\033[91m'

MirkoCovizzi:
A way to enable this functionality is by calling the Windows function `system` passing the command we want, in this case the const char* command `"color"`. Then, every print with ansi escape codes should work normally. At least it does in Python (just tested).

https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/system-wsystem?view=vs-2019

davidlattimore:
I'm definitely happy to provide assistance if you'd like to try to fix this. Thanks! I'll make some assumptions, but if anything isn't sufficiently clear, or you'd like more detail, please ask.

If you start by doing a git clone of the evcxr repository, then cargo build. You should then be able to run target/debug/evcxr (possibly with \ on windows I guess). There is one recently merged PR that might be relevant - 36ab39320fa946532e0574c7af00345049b16888 - it's not yet in a release.

MirkoCovizzi:
Thank you!

MirkoCovizzi:
[...]

PR #115, opened 2020-04-15
Fix unicode color console output on Windows

MirkoCovizzi:
This solves the issue #114

### DIFF
--- a/evcxr_repl/src/bin/evcxr.rs
+++ b/evcxr_repl/src/bin/evcxr.rs
@@ -52,6 +52,9 @@ fn send_output<T: io::Write + Send + 'static>(
 
 impl Repl {
     fn new(ide_mode: bool) -> Result<Repl, Error> {
+        #[cfg(windows)]
+        let _ret = control::set_virtual_terminal(true);
+
         let (command_context, outputs) = CommandContext::new()?;
         send_output(outputs.stdout, io::stdout(), None);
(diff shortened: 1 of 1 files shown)


MirkoCovizzi:
@googlebot I signed it!

evcxr_repl/src/bin/evcxr.rs
```diff
@@ -52,6 +52,7 @@ fn send_output<T: io::Write + Send + 'static>(
 
 impl Repl {
     fn new(ide_mode: bool) -> Result<Repl, Error> {
+        let _ret = control::set_virtual_terminal(true);
```
davidlattimore, line 55: Thanks for the fix! The [docs](https://docs.rs/colored/1.9.3/x86_64-pc-windows-msvc/colored/control/fn.set_virtual_terminal.html) say this  function is only available on windows, so you probably need to use `#[cfg(windows)]` to make sure this line is excluded on other platforms.

MirkoCovizzi: You are right! I knew I was missing something, thank you for the heads up!

MirkoCovizzi: @davidlattimore Is it fine if I use the cfg! macro? Otherwise I think I would need to make a separate function for it. Which approach would you prefer?

PR #120, opened 2020-05-25
Added explicit image import to sample; adapted code to match recent version.

mkulke:
The code in the binder sample which was using the image crate was not working anymore, since the API changed. I added an explicit dep statement and fixed the code according to the current version.


davidlattimore:
Thanks for the fix!

PR #121, opened 2020-05-25
Fix loading crates with a hyphen when using `extern crate`

### LINKED ISSUE #61: :dep names with dashes should be replaced with underscores
Looks like there was a fix for dependencies in https://github.com/google/evcxr/commit/38a948513088ed25e214c4f3e6e495a22428c5cf

but I think Im still having trouble with top level :dep that have hyphens
```
:dep generic_array = { git = "https://github.com/fizyk20/generic-array"}
extern crate generic_array;
```
```
error: no matching package named `generic_array` found
```

Ma27:
As of RFC0940[1], crates shouldn't contain any hyphens anymore. This
will cause a syntax error in Rust, however `cargo` is supposed to
internally replace hyphens with underscores.

While this is not a problem when using the `:dep`-functionality, this
will fail when using the `extern crate`-hook of the REPL since it fails
with a syntax error when using a dash, but can't find the crate if it's
specified with an underscore.

Resolves #61

[1] https://rust-lang.github.io/rfcs/0940-hyphens-considered-harmful.html

### DIFF
--- a/evcxr/src/eval_context.rs
+++ b/evcxr/src/eval_context.rs
@@ -21,6 +21,7 @@ use crate::item;
 use crate::module::{Module, SoFile};
 use crate::runtime;
 use crate::statement_splitter;
+use crates_io_api::SyncClient;
 use regex::Regex;
 use std;
 use std::collections::{HashMap, HashSet};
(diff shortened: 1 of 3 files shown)


evcxr/src/eval_context.rs
```diff
@@ -225,7 +225,7 @@ impl EvalContext {
                                 .external_deps
                                 .entry(crate_name.clone())
                                 .or_insert_with(|| {
-                                    ExternalCrate::new(crate_name.clone(), "\"*\"".to_owned())
+                                    ExternalCrate::new(crate_name.replace("_", "-").clone(), "\"*\"".to_owned())
```
bjorn3, line 228: What if both the rustc crate name and crates.io package name are using `_`, like [`rand_core`](https://crates.io/crates/rand_core)?

Ma27: Ahh, I wrongly interpreted `_` and `-` as exchangeable which isn't the case. Will push a fix for this soon, thanks!

Ma27: So, interestingly even the crates.io-API doesn't make a difference between hyphens and underscores, but the `Cargo.toml` does. I updated the patch to check for the actual crate's name if an underscore exists. Thanks for your feedback!

evcxr/src/eval_context.rs
```diff
@@ -219,12 +220,22 @@ impl EvalContext {
...
+                        let mut crate_name = ident.to_string();
                         if !self.dependency_lib_names()?.contains(&crate_name) {
                             self.state
                                 .external_deps
                                 .entry(crate_name.clone())
                                 .or_insert_with(|| {
+                                    // For RFC0940
+                                    if crate_name.contains("_") {
+                                        let crates_client = SyncClient
+                                            ::new("evcxr", std::time::Duration::from_millis(500)).unwrap();
```
bjorn3, line 232: `cargo-edit` has an implementation of this that uses the local registry instead of the crates.io api: [`fuzzy_query_registry_index`](https://github.com/killercup/cargo-edit/blob/f7c89623813023a604fad50512a5dc12cd985548/src/fetch.rs#L263-L297)

Ma27:
Don't have the time to continue working on this, sorry! Using `:dep` is a fair workaround IMHO.

PR #125, opened 2020-07-08
Stop people crashing their computers

Baughn:
Twice, in fact.

I'm working on a Discord bot version of evcxr, for helping a friend learn Rust. That's probably not worth upstreaming; it won't be secure in the slightest.

### DIFF
--- a/evcxr/README.md
+++ b/evcxr/README.md
@@ -16,6 +16,8 @@ context.eval("s.push_str(\" World\");")?;
 context.eval("println!(\"{}\", s);")?;
 ```
 
+You must call ```evcxr::runtime_hook()``` at the top of main, otherwise the library becomes a fork-bomb.
+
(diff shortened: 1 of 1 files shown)


Baughn:
The presubmit failure is unrelated; this is a documentation change only.

davidlattimore:
Thanks for the contribution and sorry about the fork bomb. If I get the chance, I'll set an environment variable when launching the subprocess, then error if that environment variable is already set when we go to launch another subprocess.

Interesting idea having a Discord bot!

Baughn:
It's more annoying than I might have hoped, but I was able to hack together something that worked in a couple of hours.

Really, the biggest issue is that aforementioned lack of security. Absolutely nothing stops a user from calling std::fs. That's usually a feature, but not in this case...

PR #127, opened 2020-07-23
display_errors() does not work properly with multi line inputs.

aminroosta:
The right side is master, the left side is this PR's changes.

image

For multi-line inputs, printing the span lines, makes it more readable - I think :)


aminroosta:
@googlebot I signed it!

davidlattimore:
Nice. That looks much better. Thanks for doing that @aminroosta !

Issue #102, opened 2020-02-05
Suggestion: Add rustanalyzer integration


CGMossa:
This would solidify the REPL-experience on Jupyter notebooks.

I don't know how this could be achieved, or what the necessary steps. If someone knows, then
we could initiate a conversation here.

If this is not appropriate, feel free to close this.

davidlattimore:
There's #3 for adding tab completion. Out of curiosity, what would be the other top feature requests? It's possible that each feature of rust analyzer might require separate work to integrate.

I have previously looked at trying to use rust analyzer to do tab completion. I had trouble. I think perhaps I needed to do something else to get it to load the standard library because the completions it was returning didn't seem to include any standard library stuff.

I'm also not sure if the approach that I was trying to take was the right one. I was trying to use rust analyzer directly as a rust library, but perhaps I should have been starting the rust analyzer language server, then connecting to that to make requests. I do like the idea of just using it as a library though, since then it could just be built into the evcxr binary and wouldn't need a separate installation step. OTOH, if it used the language server, it would make it easier for people to upgrade it separately without me doing new evcxr releases.

bjorn3:
Rust analyzer supports signature help, type info for variables, showing (syntax) errors while typing, goto definition/usage, several assists (inlining/extracting variables, adding imports for not yet imported things, adding derives), semantic syntax highlighting (showing which variables are mutable, rainbow highlighting) and more.

CGMossa:
@davidlattimore I don't know what would be the most sensible. I'd vote for the server-approach, mainly because I assume that there would be features that would lean into this kind of setup, as rustanalyzer is used like that with vscode, emacs, etc. But there is an advantage to the baked-in features. Maybe both?

I'm very happy that this wasn't outright dismissed, thank you!

davidlattimore:
It looks like language server protocol (LSP) support for Jupyter Lab has come quite a long way since last I looked. See [krassowski/jupyterlab-lsp](https://github.com/krassowski/jupyterlab-lsp).

It wouldn't help the REPL, but it would be awesome to have something like that working with the Jupyter Kernel.

I don't think it would work to have jupyterlab-lsp connect directly to a Rust language server, since it wouldn't handle :dep commands, existing variables, writing statements outside of functions etc. So Evcxr would need to act as an LSP proxy, altering the requests as they pass through and the responses on the way back.

jgarvin:
[...]

PR #129, opened 2020-07-24
naive-tab-completion

aminroosta:
In absence of rls (#3) or rust-analyzer integration (#102), I think even a basic, naive tab completion is helpful.

```
let foo = "foobar";
dbg!(f<cursor>
         ^ with this PR pressing tab, cycles through foo and foobar
```

It's based on history.
And it only matches the longest prefix.


davidlattimore:
Thanks

PR #134, opened 2020-08-06
evcxr_repl: Don't override prompt highlighting

tjg1:
The colour for the prompt already gets set here:

https://github.com/google/evcxr/blob/master/evcxr_repl/src/bin/evcxr.rs#L234

and this makes customising the prompt colour impossible when integrating the REPL into another project

### DIFF
--- a/evcxr_repl/src/repl.rs
+++ b/evcxr_repl/src/repl.rs
@@ -68,6 +68,11 @@ impl Completer for EvcxrRustylineHelper {
     }
 }
 
+/*
+ * This single line.
+ * Copyright (c) rd <evcxr@megane.space>
+ * License: WTFPL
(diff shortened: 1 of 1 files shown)


tjg1:
This is a trivial change and I'd prefer not to sign the Google CLA.

If that's a problem, please disregard this pull request.

Thanks.

davidlattimore:
Thanks for the contribution. If it were up to me, I wouldn't have a CLA at all. Unfortunately I'm not allowed to accept PRs without the CLA

davidlattimore:
Sorry. Unfortunately, I still can't merge this. Code under different licenses is permitted, but it needs to go in separate third_party directory. That wouldn't really work here.

PR #136, opened 2020-09-15
added quit command

komi1230:
Hi, thank you for the awesome project !

I felt it would be more friendly if `evcxr` has `:quit` command.  
And I added this command.


komi1230:
@googlebot I signed it!

davidlattimore:
Good idea. I'm so used to pressing ^D to exit from shells that I never noticed its absence. Thanks for the PR!

PR #155, opened 2021-02-20
Add build-essential as installation requirement

okdewit:
Without the Ubuntu build-essential package, zmq-sys will not be able to compile as it can not find `/usr/bin/cc` or whatever CMAKE_C_COMPILER is supposed to be set to.

Most developers will have build-essential already installed, but when setting this up on a new machine it might be missing.


okdewit:
@googlebot I signed it!

davidlattimore:
Thanks @okdewit !

Issue #161, opened 2021-03-22
Relative paths don't work


CinusMinus:
When using a local crate with `:dep local-crate = { path = "../local-crate" }` the import fails with a message like this:
```
>> :dep local-crate = { path = "../local-crate" }
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ 
error: failed to get `local-crate`

Caused by:
  failed to load source for dependency `local-crate`

Caused by:
  Unable to update /tmp/local-crate

Caused by:
  failed to read `/tmp/local-crate/Cargo.toml`

Caused by:
  No such file or directory (os error 2)
```
This problem has occured before in #14 and was originally fixed in e2d06b7 but it seems it reemerged somehow. The generated Cargo.toml does not make the relative paths absolute:

```toml
    [package]
    name = "evcxr_dummy_validate_dep"
    version = "0.0.1"
    edition = "2018"

    [lib]
    path = "lib.rs"

    [dependencies]
    local-crate = { path = "../local-crate" }    
```

wrwg:
Unfortunately, this is kind of a serious blocker to share notebooks. I have some analysis I want to do and the relating notebooks I like to check into a repo and others can then replay this. But they will need to edit the `:dep` directives because I can't use relative ones.

davidlattimore:
Thanks for the feedbak @wrwg. Does @CinusMinus 's fix above solve this issue for you? You should be able to test it by installing evcxr_jupyter via the following command:

```sh
cargo install --force --git https://github.com/google/evcxr.git evcxr_jupyter
```

davidlattimore:
Just wondering if I just need to do a release, or if some more fixing is needed

wrwg:
Thanks @davidlattimore amd @CinusMinus. Yes the fix works for me ;-) Releasing would be great.

PR #162, opened 2021-03-22
Absolute paths in dependency validation

CinusMinus:
Fixes the problems described in #161 to make sure local packages can be identified by relative paths


davidlattimore:
Thanks for the fix @CinusMinus !

At some point, I'll have to add a test to prevent myself regressing this again :)

PR #171, opened 2021-05-03
Allow optionally (on by default) using mimalloc as a global allocator in binaries.

thomcc:
On my macOS machine, this brings down the startup time of `evcxr_repl` (mostly spent initializing rust-analyzer) from ~3.4-3.8s to ~1.5-1.8s(!!!).

I'd expect other OSes to be less dramatic (userspace allocation on macOS has... some issues), but mimalloc is insanely fast, and this seems to help a great deal.

I don't use jupyter, and so I am just assuming this is worth doing for evcxr_jupyter too — I can't really say how much this helps. I figure there's likely no reason not to do it for both of the binary crates. I set it as optional, in case someone hits issues and needs to disable it — weird platform issues, for example.

I also disabled the default features, since they're security checks (guard pages, heap verification, and the like). My thoughts on that are https://github.com/purpleprotocol/mimalloc_rust/issues/39 — We're already using a memory safe language, so IMO that stuff should be opt-in. Oh well, easy enough for us to turn off, it's not like evcxr has a ton of pointer manipulation or anything anyway.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -558,6 +558,7 @@ dependencies = [
  "hex",
  "hmac",
  "json",
+ "mimalloc",
  "sha2",
  "unicode-segmentation",
  "uuid",
@@ -570,6 +571,7 @@ version = "0.9.0"
 dependencies = [
  "colored",
  "evcxr",
+ "mimalloc",
  "once_cell",
  "regex",
  "rustyline",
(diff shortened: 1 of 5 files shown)


davidlattimore:
Thanks @thomcc! I tried it on Linux and it doesn't seem to make any difference, so I might end up disabling it there just to save the extra dependency. That's great that it has such an effect on Mac though. Does it speed up running code, or just startup? Either way, it's a great improvement.

thomcc:
> Does it speed up running code, or just startup? Either way, it's a great improvement.

It's harder to say, but probably. I'm a bit surprised it's not a win on linux (my experinece is that mimalloc outperforms ptmalloc almost always), but it's certainly not inconceivable.

PR #174, opened 2021-05-07
Initialize the repl's CommandContext in a background thread

thomcc:
This allows bringing up the repl immediately, and *vastly* improves the perceived startup time. (It feels instant now). This allows me to bring it up, start typing/thinking about what I'm writing while the slow init (mostly of r-a) continue. If you submit the prompt before init finishes, it will join, blocking until the thread completes. The same will happen if you trigger completions — anything that requires `lock()`ing the `CommandContext`'s mutex.

This also changes `:load_config` to accept an optional `--quiet` flag, which prevents it from `println!`ing, as doing so would clobber the command line in progress. I did this rather than delete those lines just because it's nice for if the user manually reloads the files, and IDK, maybe you want them in the jupyter version.

---

The abstraction used for this (`BgInitMutex`) is a bit janky, but not that complex. I tried to document  it pretty well, even if it isn't that complex, confusion in concurrent code can lead to a lot of bugs.

The name "`BgInitMutex`" is kinda bad, I avoided using the term `Async` in the name mostly because it doesn't use async/await, and it behaves differently (Rust async/await is lazy (doing nothing unless polled), whereas this is (necessarily) eager). That said, the stuff in the `evcxr_repl` library crate isn't really public, so a bad name doesn't really matter that much.

---

Eventually, I'd like to move more operations to a background thread (when you add a `:dep`, the next compile can be quite slow, and gives no progress indication), but for now this tries to be a simple change that doesn't make unrelated code more complex or have to worry about this change at all — it just acts like a different `Mutex<T>`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -572,6 +572,7 @@ dependencies = [
  "evcxr",
  "mimalloc",
  "once_cell",
+ "parking_lot",
  "regex",
  "rustyline",
  "structopt",
--- a/evcxr/src/command_context.rs
+++ b/evcxr/src/command_context.rs
@@ -219,12 +219,14 @@ Panic detected. Here's some useful information if you're filing a bug report.
         Ok(completions)
     }
 
-    fn load_config(&mut self) -> Result<EvalOutputs, Error> {
+    fn load_config(&mut self, quiet: bool) -> Result<EvalOutputs, Error> {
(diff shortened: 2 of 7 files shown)


davidlattimore:
Nice, that's awesome! I like that you've even put thought into how crashes will be reported if something goes wrong while initializing.

PR #206, opened 2021-12-16
Add support for linking with mold

wseaton:
This patch adds support for using `mold` as the linker for `evcxr` by injecting it in-front of the `cargo` compile commands (if configured), the rationale for that is discussed here in: https://github.com/rui314/mold#how-to-use

Tested mostly locally in JupyterLab and seems to work for me, let me know if I'm missing anything!

EDIT: Extracting the mold `.comment` from the generated binary: 
```sh
.tmpVO1W2R/target/debug 
❯ readelf -p .comment libctx.so 

String dump of section '.comment':
  [     0]  GCC: (GNU) 11.2.1 20211203 (Red Hat 11.2.1-7)
  [    2e]  mold 1.0.0 (ed9924895d9b9584106791247596677db8113528; compatible with GNU ld and GNU gold)

```

### DIFF
--- a/COMMON.md
+++ b/COMMON.md
@@ -187,7 +187,7 @@ Here is a complete list of the configuration options you can set to customize yo
 * `:fmt [format]`     Set output formatter (default: `{:?}`). 
 * `:efmt [format]`    Set the formatter for errors returned by `?`
 * `:sccache [0|1]`    Set whether to use sccache.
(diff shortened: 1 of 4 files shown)


wseaton:
Signed the CLA as an individual.

davidlattimore:
Thanks for the contribution!

PR #236, opened 2022-08-11
Use ariadne for rendering diagnostics

HKalbasi:
Before:

In REPL:
image

in jupyter notebook:
image

After:

in REPL:
image

I didn't managed to make it working in jupyter notebook.

This PR is just a proof of concept, for giving feedback. It doesn't actually work for anything other than my example (I added a `+20` offset for spans somewhere to make it working) and it should show helps and notes using ariadne as well (the currently rendered help is from old code). [Ariadne has a very easy to use api](https://github.com/zesterer/ariadne/tree/main/examples), it gets a list of files with their sources (commands and cells in our use case) and a list of spans (declared with byte ranges) with messages and colors, then it will automatically render everything in the right place. So someone that know more than me (= more than nothing) about `evcxr` can do it the right way fairly easily. But I can also make it with some mentoring if needed.

Rust is famous for its great compile messages. Lets make them great in `evcxr` as well!

### DIFF
--- a/evcxr/src/errors.rs
+++ b/evcxr/src/errors.rs
@@ -36,6 +36,7 @@ pub struct CompilationError {
     pub json: JsonValue,
     pub(crate) code_origins: Vec<CodeKind>,
     spanned_messages: Vec<SpannedMessage>,
+    spanned_helps: Vec<SpannedMessage>,
     level: String,
 }
 
@@ -51,7 +52,9 @@ fn span_to_byte_range(source: &str, span: &Span) -> Range<usize> {
             .take(line_number - 1)
             .map(|x| x.len())
             .sum::<usize>()
-            + column + line_number - 2
+            + column
+            + line_number
+            - 2
     }
     line_and_number_to_byte_offset(source, span.start_line, span.start_column)
         ..line_and_number_to_byte_offset(source, span.end_line, span.end_column)
@@ -60,16 +63,16 @@ fn span_to_byte_range(source: &str, span: &Span) -> Range<usize> {
 impl CompilationError {
     pub fn build_report(
         &self,
-        command_history: &[(&'static str, &'static str)],
+        file_name: String,
+        source: String,
         theme: Theme,
-    ) -> Option<Report<(&'static str, Range<usize>)>> {
+    ) -> Option<Report<(String, Range<usize>)>> {
         let error = self;
-        let (file_name, source) = command_history.last().unwrap();
         if !source.is_ascii() {
             return None;
         }
         let mut builder =
-            Report::build(ReportKind::Error, *file_name, 0).with_message(&error.message());
+            Report::build(ReportKind::Error, file_name.clone(), 0).with_message(&error.message());
         let mut next_color = {
             let mut colors = ColorGenerator::new();
             move || {
@@ -86,10 +89,17 @@ impl CompilationError {
         if let Some(code) = error.code() {
             builder = builder.with_code(code);
         }
-        for spanned_message in error.spanned_messages() {
+        for spanned_message in error
+            .spanned_messages()
+            .iter()
+            .chain(error.help_spanned().iter())
+        {
             if let Some(span) = &spanned_message.span {
+                if spanned_message.label == "" {
+                    continue;
+                }
                 builder = builder.with_label(
-                    Label::new((*file_name, span_to_byte_range(source, span)))
+                    Label::new((file_name.clone(), span_to_byte_range(&source, span)))
                         .with_message(&spanned_message.label)
                         .with_color(next_color())
                         .with_order(10),
@@ -188,6 +198,16 @@ impl CompilationError {
 
         Some(CompilationError {
             spanned_messages: build_spanned_messages(&json, code_block),
+            spanned_helps: {
+                if let JsonValue::Array(children) = &json["children"] {
+                    children
+                        .iter()
+                        .flat_map(|x| build_spanned_messages(x, code_block))
+                        .collect()
+                } else {
+                    vec![]
+                }
+            },
             message,
             level: json["level"].as_str().unwrap_or("").to_owned(),
             json,
@@ -215,6 +235,7 @@ impl CompilationError {
     ) -> CompilationError {
         CompilationError {
             spanned_messages: vec![spanned_message],
+            spanned_helps: vec![],
             message,
             json: JsonValue::Null,
             code_origins: vec![segment.kind.clone()],
@@ -280,6 +301,10 @@ impl CompilationError {
 
     pub fn level(&self) -> &str {
         &self.level
+    }
+
+    pub fn help_spanned(&self) -> &[SpannedMessage] {
+        &self.spanned_helps
     }
 
     pub fn help(&self) -> Vec<String> {
@@ -365,9 +390,25 @@ fn sanitize_message(message: &str) -> String {
 
 fn build_spanned_messages(json: &JsonValue, code_block: &CodeBlock) -> Vec<SpannedMessage> {
     let mut output_spans = Vec::new();
+    let mut only_one_span = false;
+    let level_label: Option<String> = (|| {
+        let level = json["level"].as_str()?;
+        if level != "error" {
+            // We can't handle helps and notes with multiple spans currently
+            only_one_span = true;
+        }
+        let message = json["message"].as_str()?;
+        Some(format!("{level}: {message}"))
+    })();
     if let JsonValue::Array(spans) = &json["spans"] {
-        for span_json in spans {
-            output_spans.push(SpannedMessage::from_json(span_json, code_block));
+        if !only_one_span || spans.len() == 1 {
+            for span_json in spans {
+                output_spans.push(SpannedMessage::from_json(
+                    span_json,
+                    code_block,
+                    level_label.clone(),
+                ));
+            }
         }
     }
     if output_spans.iter().any(|s| s.span.is_some()) {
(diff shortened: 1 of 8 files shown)


bjorn3:
Wouldn't it be an idea to just reuse rustc's rendered diagnostics? Those are included in the json error messages.

HKalbasi:
I don't think it is easier. It has wrong line numbers, it may show some of the generated codes, and there is some processing on the error message (for example adding some `evcxr` specific suggestions). If there were no ready `ariadne`, that was the way, but now I think this is better. Do you have a specific reason to reuse rustc diagnostics?

HKalbasi:
@davidlattimore what is your opinion on this? I'm interested in putting more work on this if you think this in the right direction.

davidlattimore:
Hi @HKalbasi - sorry for the delayed response. I started looking before, but must have gotten distracted and forgot to come back to it.

Do you know whether it works on Windows and Mac? At the moment, evcxr_repl uses the `colored` crate, which takes care of some of the platform differences. I don't have Windows or Mac, so I can't test how `ariadne` works there.

The rendering of the errors does look nice. The dependency looks lightweight. Assuming it works on Windows and Mac, or you can provide a fallback to the old behavior on those platforms, it seems reasonable to me. Having someone pay some extra attention to error display sounds great - thanks :)

HKalbasi:
> Hi @HKalbasi - sorry for the delayed response. I started looking before, but must have gotten distracted and forgot to come back to it.

That's fine. Thanks for maintaining this!

> Do you know whether it works on Windows and Mac? At the moment, evcxr_repl uses the colored crate, which takes care of some of the platform differences. I don't have Windows or Mac, so I can't test how ariadne works there.

I don't have windows or mac either, but I made some research. If I understand correctly, `ariadne` uses `yansi` instead of `colored` which only supports `ANSI` terminals. `ANSI` is supported on major terminals on Linux, official terminal in mac, windows terminal and latest versions of powershell. But windows cmd needs explicit request from the application, and it only works on windows 10. By including this code somewhere:
 ```
if cfg!(windows) && !Paint::enable_windows_ascii() {
    Paint::disable()
}
```
We will request `ANSI` support on windows and disable coloring if it is not available (on windows 8 and lower). Is it fine?

I also have some questions about how make it working in jupyter (I think it should work as is but I'm not able to use my build as the kernel) and how finding exact byte offset, which I will show on the code.

File: evcxr/src/errors.rs [hunk 1]
```diff
@@ -52,18 +52,41 @@ fn spans_in_local_source(span: &JsonValue) -> Option<&JsonValue> {
...
+        for x in &code_block.segments {
+            if x.code.len() > be {
+                break;
+            }
+            if matches!(x.kind, CodeKind::OriginalUserCode(_)) {
+                break;
+            }
+            be -= x.code.len();
+        }
+        (code_origins, (bs, be))
```
HKalbasi, line 86: This code tries to covert the byte offset returned by cargo, to the byte offset inside user code. I currently tried to decrease size of code segments other than `CodeKind::OriginalUserCode` that may appear before user's code. What should I do here?

HKalbasi: It seems it works on single statement blocks. If I remove `+20`, the query `2 + 3.` would work correctly:
image
But not on multiple statements blocks:
image

HKalbasi: Nevermind I think I fixed it.

davidlattimore (commented):
Colour on recent versions of Windows and fallback to non-colour on older versions sounds reasonable to me. Thanks for investigating that.

To use your build of `evcxr_jupyter`, run:

```
cargo build -p evcxr_jupyter
./target/debug/evcxr_jupyter --install
```

Then restart jupyter notebook.

If I paste in the original block of code that you had above, the spans appear to be wrong.

```rust
use std::thread;

let mut x = 0;
let join_handle = thread::spawn(|| {
  x += 2;
  println!("{x}");
});
x += 2;
println!("{x}");
join_handle.join();
println!("{x}");
```

```
[E0373] Error: closure may outlive the current function, but it borrows `x`, which is owned by the current function
   ╭─[command_0:1:1]
   │
 4 │ let join_handle = thread::spawn(|| {
   ·               ─┬     ┬  
   ·                ╰──────── may outlive borrowed value `x`
   ·                      │  
   ·                      ╰── `x` is borrowed here
───╯
help: to force the closure to take ownership of `x` (and any other referenced variables), use the `move` keyword

move
[E0503] Error: cannot use `x` because it was mutably borrowed
   ╭─[command_0:1:1]
   │
 4 │ ╭─▶ let join_handle = thread::spawn(|| {
   · │                 ─┬     ┬  
   · │                  ╰──────── borrow of `x` occurs here
   · │                        │  
   · │                        ╰── borrow occurs due to use of `x` in closure
   ⋮ ⋮   
 6 │ ├─▶   println!("{x}");
   · │        ───┬──          
   · ╰──────────────────────── argument requires that `x` is borrowed for `'static`
   ·             │            
   ·             ╰──────────── use of borrowed `x`
```

evcxr_repl/src/bin/evcxr.rs
```diff
@@ -120,51 +127,42 @@ impl Repl {
...
+                let mut builder =
+                    Report::build(ReportKind::Error, self.command_history.last().unwrap().0, 0)
+                        .with_message(&error.message());
+                let mut colors = ColorGenerator::new();
+                if let Some(code) = error.code() {
+                    builder = builder.with_code(code);
+                }
                 for spanned_message in error.spanned_messages() {
                     if let Some(span) = &spanned_message.span {
-                        let mut start_column = character_column_to_grapheme_number(
```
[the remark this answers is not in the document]
HKalbasi: Not only grapheme clusters are not supported (which I think it is something on `ariadne` side) even non ascii unicode characters will break the offsets (which it should be possible, but I didn't figured it out). I restored that codes and made them fallback for non-ascii source codes, for now:
image
Is it fine?

HKalbasi:
> To use your build of evcxr_jupyter, run:

It's working now, thanks! It is the result on vscode's jupyter:
image
But it is not great on original jupyter with white theme:
image
Do you know how I can force it to use black for background and white for text?

> If I paste in the original block of code that you had above, the spans appear to be wrong.

This is bad. It's related to the additional new line in that code, which we don't have in the code segments, so the for that computes span offsets fails in this case. Is it possible to keep whitespaces between semicolons inside code segments?

davidlattimore:
I'm not sure how to change Jupyter Notebook to have a black background. My guess is it might require some kind of extension. Does ariadne support a way to use colours that look OK on white?

> > If I paste in the original block of code that you had above, the spans appear to be wrong.
> This is bad. It's related to the additional new line in that code, which we don't have in the code segments, so the for that computes span offsets fails in this case. Is it possible to keep whitespaces between semicolons inside code segments?

I'm not sure the approach of using `byte_start` and `byte_end` from the JSON will work. Evcxr has mechanisms in place to track where each line comes from, be it from code that the user wrote, or code that was generated. It the figures out the line number and column for the start / end of a span. I'd suggest trying to use the existing `start_line`, `start_column` on `Span` rather than adding `byte_start` etc. If you need byte offsets, it should be possible to derive them from the line and column offsets.

HKalbasi:
> Does ariadne support a way to use colours that look OK on white?

I didn't find such option, but manually inverted ariadne colors on jupyter, so it is now this:
image
In this way it will break on jupyters with dark themes and vscode's jupyter. We either need to detect user's theme, or we should force black (or white) background everywhere somehow. I tried to use ANSI to change background color, but result was poor. Any suggestion?

> If you need byte offsets, it should be possible to derive them from the line and column offsets.

Thanks! I used the line and column offsets for computing byte offsets and now it seems spans are always correct.

evcxr_jupyter/src/core.rs
```diff
@@ -162,6 +166,14 @@ impl Server {
             *self.latest_execution_request.lock().unwrap() = Some(message.clone());
             let src = message.code();
             execution_count += 1;
+            self.command_history.push({
+                let name = format!("command_{}", execution_count);
+                let src = src.to_string();
+                (
+                    Box::leak(name.into_boxed_str()),
```
davidlattimore, line 173: Is it necessary to leak here, or could we store a `String` instead of a `&'static str`?

evcxr_jupyter/src/core.rs
```diff
@@ -36,6 +38,7 @@ pub(crate) struct Server {
     latest_execution_request: Arc<Mutex<Option<JupyterMessage>>>,
     shutdown_requested_receiver: Arc<Mutex<crossbeam_channel::Receiver<()>>>,
     shutdown_requested_sender: Arc<Mutex<crossbeam_channel::Sender<()>>>,
+    command_history: Vec<(&'static str, &'static str)>,
```
davidlattimore, line 41: Do we ever use the commands that were executed besides the most recent one? If we only ever need the most recent one, then do we need to have this in an field, or could it just be a local variable / argument in the code that needs it?

davidlattimore (commented):
Nice! That seems to work well. Thanks for your patience with getting this working and with my somewhat slow reviews. If any of the suggested changes / questions prove too hard, feel free to leave them and I can have a look after merging.

HKalbasi:
I removed the `command_history` and leaking, and also added limited support for showing helps and notes. Thanks for review!

By the way, is there a specific reason for not using `serde` and `cargo_metadata` and manually parsing the cargo result? `cargo_metadata` crate provides types and documentation for cargo json messages, and with it there is no need for codes like `if let Json::Array(x) = x["span"]`. Not related to this PR, I'm just curious.

davidlattimore:
Great work, thanks!

Regarding `cargo_metadata`, my guess is that I probably just didn't think to look for such a crate at the time I was writing it. I don't object to switching to using `cargo_metadata`.

PR #241, opened 2022-08-17
updating inconsistency in custom output description (different from c…

lisanhu:
From file: evcxr_jupyter/README.md 
Section: Custom output

last sentence in the 1st paragraph:
```
Each block starts with a line containing BEGIN_EVCXR_OUTPUT followed by the mime type, then a newline, the content then ends with a line containing EVCXR_END_CONTENT.
```

So it says block starts with BEGIN\_EVCXR\_OUTPUT, and from the code example below, it should be EVCXR\_BEGIN\_CONTENT

I haven't tested whether BEGIN_EVCXR_OUTPUT will do the work, but there's an inconsistency between the description and the example and I think it should be fixed.


davidlattimore:
Thanks for the fix @lisanhu!

PR #254, opened 2022-11-29
Document how to use the local crate

noamraph:
As a newbie, I had to look a bit to find how to use the crate I'm working on from evcxr - I wasn't sure it was even possible. I guess this is a common use case, so I think it's worthwhile to mention it explicitly.


davidlattimore:
Thanks @noamraph!

PR #256, opened 2022-12-16
Add a command that displays the type of outputs

noamraph:
I find it very useful to see the type of expressions. So I tried to add this feature to evcxr. What I did works like this:

```
>> :t
Types: true

>> 3 + 5
: i32 = 8
>> "hello"
: &str = "hello"
>> Some("hello".to_string())
: Option<String> = Some("hello")
>> vec![1, 2, 3]
: Vec<i32> = [1, 2, 3]
```

To get the type names I used `std::any::type_name()`, and added a function which only leaves the last part of paths, so it converts `core::option::Option<alloc::string::String>` into `Option<String>`, which is much easier on the eyes.

Things which I chose arbitrarily and should be decided:

* The name of the command (`:t`)
* The format (I used `: <type> = <repr>`)

A note: for shortening the name I pushed a function (`evcxr_shorten_type`) into the child. It's quite long, and has some intricate logic, because it had to be written without relying on regex. I think it would be better to send the raw type string along with the value, and shorten it at the main process using regex. If the concept is approved, I would be happy to try it. For now, it was simpler for me to avoid changing the interface between the processes and instead add the string on the child side.

What do you think?

Thanks,
Noam

### DIFF
--- a/COMMON.md
+++ b/COMMON.md
@@ -182,6 +182,7 @@ Here is a complete list of the configuration options you can set to customize yo
 
 * `:opt [level]`      Toggle/set optimization level
 * `:fmt [format]`     Set output formatter (default: `{:?}`). 
+* `:t`                Toggle printing of the type of the output
 * `:efmt [format]`    Set the formatter for errors returned by `?`
 * `:sccache [0|1]`    Set whether to use sccache.
 * `:linker [linker]`  Set/print linker. Supported: `system`, `lld`, `mold`
--- a/COMMON.md
+++ b/COMMON.md
@@ -182,7 +182,7 @@ Here is a complete list of the configuration options you can set to customize yo
 
 * `:opt [level]`      Toggle/set optimization level
 * `:fmt [format]`     Set output formatter (default: `{:?}`). 
-* `:t`                Toggle printing of the type of the output
+* `:types`            Toggle printing of the type of the output
 * `:efmt [format]`    Set the formatter for errors returned by `?`
 * `:sccache [0|1]`    Set whether to use sccache.
 * `:linker [linker]`  Set/print linker. Supported: `system`, `lld`, `mold`
--- a/evcxr/src/command_context.rs
+++ b/evcxr/src/command_context.rs
@@ -437,11 +437,11 @@ Panic detected. Here's some useful information if you're filing a bug report.
                 },
             ),
             AvailableCommand::new(
-                ":t",
+                ":types",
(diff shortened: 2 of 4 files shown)


davidlattimore:
Hi @noamraph. Thanks for the PR! Sorry about the slow reply. I was just going through my inbox and found I had overlooked this.

Do you think doing everything in the parent process would be an option? We already have rust-analyzer figuring out the types of variables, so I think it should be possible to get it to figure out the type of an expression. Then there's no need to add any additional code to the compilation unit (which has the potential to slow down each compilation slightly).

Any thoughts on a less abbreviated name? We don't currently have any single letter commands. Perhaps `:types` or `:show_types`?

noamraph:
Hi @davidlattimore, thanks for the reply! I think that `:types` is a good name. I'll see if I manage to get the type from rust analyzer, I agree that would be the best solution.

davidlattimore:
Cool. If you get stuck, do let me know and I'll see if I can help

noamraph:
@davidlattimore I tried to figure out how to get the type of the expression using rust-analyzer, and I didn't manage to do it. Rust-analyzer finds the types from `RustAnalyzer.top_level_variables()`, which gets a pre-processed code, in which the expression was already wrapped with a function that prints it. Also, the function `ContextState::apply()`, which creates the code that prints the expression, doesn't have access to the analyzer, and gets just what the user wrote, which I guess will confuse rust-analyzer. Do you have a suggestion where finding the type of the last expression could take place? 

I tend to think that using `std::any::type_name` as I did is not a bad idea. Perhaps I should just make `evcxr_shorten_type()` shorter? If we go over the string in reverse order, and when we encounter a `:` drop all characters until we reach a character which isn't alphanumeric or `:`, the function would be much shorter. Also, I should put this function in a separate constant, which will only be used if `:types` is enabled, so it won't affect the compilation times for users who are not interested in the types.

davidlattimore:
Thanks for looking into that @noamraph. I had a look and agree that it would definitely be tricky.

Provided the extra compilation code is made conditional on the types feature being enabled, I'm fine with what you propose.

noamraph:
@davidlattimore Ok, now the command is `:types` instead of `:t`, the code to shorten the type name is much shorter, and it is only compiled if types are enabled. How does it look?

davidlattimore:
Looks good to me. Thanks for you contribution @noamraph!

PR #262, opened 2023-01-09
evcxr_jupyter: Add control channel response for kernel_info_request.

AH-DL:
Jupyter server expects a message reply for "kernel_info_request" over control channel, so i just added the message response in handle_control() function.


davidlattimore:
Thanks @AH-DL! I hadn't observed a kernel_info_request on the control channel, but looking at the [spec](https://jupyter-client.readthedocs.io/en/stable/messaging.html), it seems that the control channel is supposed to be just like the shell channel, but separate, so it makes sense that it's valid. I should probably merge the two together to use the same code.

PR #266, opened 2023-02-08
Change license to dual Apache2/MIT

davidlattimore:


davidlattimore:
Since all contributors have accepted the CLA, this change can be made without needing to check with everyone. However I wouldn't make this change if I didn't think that it would generally be considered a positive change. I would have dual licensed it from the start, however it wasn't permitted at the time. Although the CLA allows this, if any contributors object to adding the MIT license, please do let me know and I'll see what I can do to remove the code that you don't want under the MIT license.

davidlattimore:
Good idea, thanks!

@davidB @aminroosta @noamraph @ibraheemdev @dima74 @thomcc @wseaton @luizirber @ZoeyR @AH-DL @HKalbasi @JohnScience @mkulke @KeenS @thomasjm - anyone have any objections to an MIT license? Feel free to respond before or after I merge this and I'll see what I can do. There's a small, but non-zero chance that I could lose the ability to merge this at any moment, so I'll probably merge this in about 12 hours from now.

PR #267, opened 2023-02-10
WIP: tee-ing  output to stdout and var

### LINKED ISSUE #224: Any verbosity option for `:dep`?
Since it often takes a long time when adding external crate dependencies by running `:dep` command(s), it'd be a helpful to know what's going on or even better to see a progress bar. Is there any way to display verbose messages when running `:dep` command?

adiSuper94:
resolves #224

### DIFF
--- a/evcxr/src/module.rs
+++ b/evcxr/src/module.rs
@@ -14,9 +14,11 @@ use crate::eval_context::ContextState;
 use once_cell::sync::OnceCell;
 use regex::Regex;
 use std::fs;
+use std::io::Read;
 use std::path::Path;
 use std::path::PathBuf;
 use std::process::Command;
+use std::process::Stdio;
 
 fn shared_object_name_from_crate_name(crate_name: &str) -> String {
     if cfg!(target_os = "macos") {
@@ -265,11 +267,47 @@ overflow-checks = true
         )
     }
 }
+fn tee(mut stream: impl Read) -> Result<String, Error> {
+    let mut out: String = String::new();
+    let mut buff: [u8; 1024] = [0; 1024];
+    // I dont like infinite loops
+    loop {
+        let bytes_read = stream.read(&mut buff)?;
+        if bytes_read == 0 {
+            return Ok(out);
+        }
+        let partial_out = std::str::from_utf8(&buff)?;
+        //TODO: maybe pass a callback here
+        print!("{partial_out}");
+        out.push_str(partial_out);
+    }
+}
 
 fn run_cargo(
     mut command: std::process::Command,
     code_block: &CodeBlock,
 ) -> Result<std::process::Output, Error> {
+    let mut child_process = match command
+        .stdout(Stdio::piped())
+        .stderr(Stdio::piped())
+        .spawn()
+    {
+        Ok(child) => child,
+        Err(err) => bail!("Error running 'cargo rustc' as a child process: {}", err),
+    };
+    let child_out = child_process.stdout.take().unwrap();
+    let child_err = child_process.stderr.take().unwrap();
+    let out = match tee(child_out) {
+        Ok(out) => out,
+        Err(err) => bail!("oouch : {}", err),
+    };
+
+    let err = match tee(child_err) {
+        Ok(out) => out,
+        Err(err) => bail!("oouch 2 :{}", err),
+    };
+
+    println!("out is : {out} \n err is : {err}");
     let cargo_output = match command.output() {
         Ok(out) => out,
         Err(err) => bail!("Error running 'cargo rustc': {}", err),
--- a/evcxr/src/module.rs
+++ b/evcxr/src/module.rs
@@ -22,7 +22,6 @@ use std::path::PathBuf;
 use std::process::Command;
 use std::process::Output;
 use std::process::Stdio;
-
 use std::thread;
 
 fn shared_object_name_from_crate_name(crate_name: &str) -> String {
@@ -272,7 +271,8 @@ overflow-checks = true
         )
     }
 }
-fn tee(mut stream: impl Read, callback: fn(partial_out: &String)) -> Result<String, Error> {
+
+fn tee(mut stream: impl Read, callback: impl Fn(&str)) -> Result<String, Error> {
     let mut buf_stream = BufReader::new(&mut stream);
     let mut out: String = String::new();
     let mut buff = String::new();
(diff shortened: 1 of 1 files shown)


evcxr/src/module.rs
```diff
@@ -265,11 +267,47 @@ overflow-checks = true
...
+fn tee(mut stream: impl Read) -> Result<String, Error> {
+    let mut out: String = String::new();
+    let mut buff: [u8; 1024] = [0; 1024];
+    // I dont like infinite loops
+    loop {
+        let bytes_read = stream.read(&mut buff)?;
+        if bytes_read == 0 {
+            return Ok(out);
+        }
+        let partial_out = std::str::from_utf8(&buff)?;
```
bjorn3, line 279: The partial input may be cut of in the middle of an UTF-8 character and thus fail to verify. You could try pushing everything as raw bytes to the buffer and then to have a live view scan from the last byte that was printed until the last newline (which is a single byte big and due to the way UTF-8 works doesn't occur in the middle of any other encoded character) and then use str::from_utf8 on this slice.

evcxr/src/module.rs
```diff
@@ -265,11 +267,47 @@ overflow-checks = true
...
+        Ok(child) => child,
+        Err(err) => bail!("Error running 'cargo rustc' as a child process: {}", err),
+    };
+    let child_out = child_process.stdout.take().unwrap();
+    let child_err = child_process.stderr.take().unwrap();
+    let out = match tee(child_out) {
+        Ok(out) => out,
+        Err(err) => bail!("oouch : {}", err),
+    };
+    // TODO: both streams should be processed concurrently.
```
davidlattimore, line 304: It may end up being simplest to spawn a thread to handle one of the streams. Spawning a thread feels like a somewhat heavyweight thing to do, but we're already spawning cargo, so a thread perhaps isn't much in comparison.

adiSuper94: Hmm makes sense. 
We use thread for now at least.  It's simple. 
It can be changed to some with better performance if needed.

evcxr/src/module.rs
adiSuper94: Not very sure about how likely it is for this to happen. 
But it is a good catch. Thank you.

adiSuper94:
Also it is `stderr`  which seems have have useful compilation prints. Not sure we print `stdout` to user.

adiSuper94:
I split the output parsing function because it is used is being used by `pub fn check`, and I didn't want to touch multiple functions.
I can revert the parsing function to its old state and change its call in `check` if needed.

evcxr/src/module.rs
```diff
@@ -266,27 +272,76 @@ overflow-checks = true
...
+        }
+        //TODO:Is state variable needed for the callback, incase the callback needs to buffer some info?
+        callback(&buff);
+        out.push_str(&buff);
+        buff.clear();
+    }
+}
+
+fn get_callback(code_block: &CodeBlock) -> impl Fn(&str) {
+    //Idea is to add some check on the code_block to decide if additional logging is necessary
```
davidlattimore, line 293: Sorting extra configuration information on the CodeBlock doesn't seem like the right place to me. You mention below that the function signature looks complex if you change run_cargo to take a callback, but is it actually that bad?

davidlattimore (commented):
Sorry for the slow review. I wrote some comments before and thought I'd sent them, but apparently I'm just not used to the github UI and a "pending" comment is actually what I would call a "draft" comment and hasn't actually been sent yet.

evcxr/src/module.rs
adiSuper94: `fn subfunc<F: 'static + FnMut() -> () + Send>(f: F) {`
details of it are discussed [here](https://users.rust-lang.org/t/passing-closure-received-as-argument-to-thread/55567/3).

I don't really mind adding this. But we need to create the callback(s) and pass it (we might need one callback for `stderr` and one for `stderr`) somewhere. And shouldn't we be adding the callback only when the codeblock that is executed is a  `:dep`. So how can we decide the callback without a reading the code_block.  
Without this check I see lot of logs that is unnecessary logs printed.

Also I am not clear about how the jupyter flow really works. I was not able to connect flow of control to the `run_cargo` fn that I am editing here.

davidlattimore: I think `run_cargo` should only need to accept callback for stderr. Processing stdout can continue to be handled entirely within `run_cargo` as it is now.

I think the callback would need to be passed through the following functions:
* `run_cargo`
* `Module::compile`

`Module::compile` is called from `EvalContext::try_run_statements`. One of the arguments to `try_run_statements` is `callbacks: &mut EvalCallbacks`. `EvalCallbacks` seems like a perfect place to add the callback.

The REPL currently calls `context.execute`, but could be changed to call `context.execute_with_callbacks`, which is the variant that takes an `EvalCallbacks` argument.

Regarding whether to only print stderr when a `:dep` command has been issued, I'm not sure. It might be simpler to just always print progress from the compiler. Or perhaps decide whether to print based on how long things are taking - e.g. for the first second, don't print anything, but then if we haven't finished compiling within a second print progress. But probably see how it looks first when doing it unconditionally. It might depend how chatty cargo is when doing normal compilation. You could also potentially filter out lines that you don't want to avoid them being printed.

adiSuper94: > I think `run_cargo` should only need to accept callback for stderr. Processing stdout can continue to be handled entirely within `run_cargo` as it is now.
> 
This makes sense. 👍

> I think the callback would need to be passed through the following functions:
> 
> * `run_cargo`
> * `Module::compile`
> 
> `Module::compile` is called from `EvalContext::try_run_statements`. One of the arguments to `try_run_statements` is `callbacks: &mut EvalCallbacks`. `EvalCallbacks` seems like a perfect place to add the callback.
> 
> The REPL currently calls `context.execute`, but could be changed to call `context.execute_with_callbacks`, which is the variant that takes an `EvalCallbacks` argument.
> 

I tried passing `EvalCallbacks` from newly created `Repl::execute_with_callbacks`  as you suggested.
But while passing the callback to the thread the callback I get an error saying `cannot be shared between threads safely`
error

I am trying to see how I can get this error fixed.

davidlattimore: If you wrap the callback in `Arc<Mutex<..>>` you should be able to share it between threads. For example:

```rust
use std::sync::Arc;
use std::sync::Mutex;

struct Foo {
    callback: Arc<Mutex<dyn FnMut() -> i32 + Send>>,
}

fn main() {
    let f = Foo {
        callback: Arc::new(Mutex::new(|| 42)),
    };
    let t = std::thread::spawn(move || {
        let callback = &mut *f.callback.lock().unwrap();
        let v = (callback)();
        println!("{v}");
    });
    t.join().unwrap();
}
```

It's fine to pass `EvalCallbacks` into `run_cargo`, but I'd probably suggest avoiding passing it to the separate thread, because then you'd need to wrap the whole of `EvalCallbacks` in `Arc<Mutex<...>>` rather than just the callback.

If your callback ends up not needing any state, then it can probably be a `Fn` rather than a `FnMut`. If that's the case, then you probably don't need the `Mutex`, just the `Arc`. e.g.:

```rust
use std::sync::Arc;

struct Foo {
    callback: Arc<dyn Fn() -> i32 + Send + Sync>,
}

fn main() {
    let f = Foo {
        callback: Arc::new(|| 42),
    };
    let t = std::thread::spawn(move || {
        let callback = &*f.callback;
        let v = (callback)();
        println!("{v}");
    });
    t.join().unwrap();
}
```

davidlattimore:
Thanks for your work on this @adiSuper94! Since @d86leader finished this off, I'll close this now, but I'll be sure to mention both of you in the release notes

adiSuper94:
@davidlattimore sorry for not promptly finishing off this task :( . 
I got caught up with with uni course work.

PR #270, opened 2023-02-17
Set `net.offline` in .cargo/config.toml according to value of :offline

### LINKED ISSUE #265: offline mode and rust-analyzer errors on loading multiple dependencies
On a machine without internet access. `:offline 1` works when there is only 1 `:dep some_crate` loaded. With multiple `:dep`s, it seems rust analyzer's `ra_ap_project_model` dies on 
```
>> :offline 1
Offline mode: true
>> :dep csv     // ok here!
>> :dep rand
Failed to read Cargo metadata from Cargo.toml file <tmp>/.tmpTQ65C4/Cargo.toml, Some(Version { major: 1, minor: 68, patch: 0, pre: Prerelease("nightly") })
```
A single crate is fine. I tracked it down to `CargoWorkspace::fetch_metadata(` but can't seem to find any more hints. There's an open issue on RA regarding being truly offline (https://github.com/rust-lang/rust-analyzer/issues/12499) .

How is `:offline 1` get passed to RA? I checked the temporary `Cargo.toml` file and don't see `net.offline = true`. Do we need to add that so RA gets it?

kting28:
- Without this, rust analyzer's cargo operations (e.g. cargo meta) still tries to access the network.
- Note - `:offline 1 or 0` does not update the .config right away. It's written together when `Cargo.toml` is updated.
- Fixes #265

### DIFF
--- a/evcxr/src/eval_context.rs
+++ b/evcxr/src/eval_context.rs
@@ -666,6 +666,7 @@ impl EvalContext {
 
     pub(crate) fn write_cargo_toml(&self, state: &ContextState) -> Result<()> {
         self.module.write_cargo_toml(state)?;
+        self.module.write_config_toml(state)?;
         Ok(())
     }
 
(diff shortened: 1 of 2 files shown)


davidlattimore:
Thanks!

PR #271, opened 2023-02-18
Fix dependency path not being TOML-escaped

### LINKED ISSUE #62: Relative path dependency on Windows not working
I'm having trouble loading a relative path dependency on Windows.  The cell looks like this:

```
:dep num = "^0.2.0"
:dep num-complex = "^0.2.3"
:dep organic-snout = { path = ".." }
:dep plotters = { git = "https://github.com/38/plotters", default_features = false, features = ["evcxr"] }
```

And the output is:

```
Compilation failed, but no parsable errors were found. STDERR:
error: failed to parse manifest at `C:\Users\ANDREW~1\AppData\Local\Temp\.tmpxYNfAx\Cargo.toml`

Caused by:
  could not parse input as TOML

Caused by:
  invalid escape character in string: `C` at line 25

STDOUT:
```

I've verified that the organic-snout dependency is the one causing the problem. If I comment it out, the cell evaluates cleanly.

If I go and look at the `Cargo.toml` that the error message refers to, it looks like this:

```
[package]
name = "ctx"
version = "1.0.0"
edition = "2018"

[lib]
crate-type = ["cdylib"]

[profile.dev]
opt-level = 2
debug = false
rpath = true
lto = false
debug-assertions = true
codegen-units = 16
panic = 'unwind'
incremental = true
overflow-checks = true

[dependencies]
num = "^0.2.0"
num-complex = "^0.2.3"
plotters = { git = "https://github.com/38/plotters", default_features = false, features = ["evcxr"] }
organic-snout = { path = "\\?\C:\Users\Andrew Watts\Projects\organic-snout" }
```

I'm not sure what that goofiness is at the beginning of the path, but clearly Cargo is confused by it, too.

Marcono1234:
Fixes #62

I have not used a TOML parser and generator for this to keep the changes small; but using a TOML parser and generator for this might be the cleaner solution in the long term.

### DIFF
--- a/evcxr/src/crate_config.rs
+++ b/evcxr/src/crate_config.rs
@@ -36,7 +36,7 @@ fn make_paths_absolute(config: String) -> Result<String, Error> {
                 Ok(path) => {
                     return Ok(captures[1].to_owned()
                         + "path = \""
-                        + &path.to_string_lossy()
+                        + &escape_toml_string(&path.to_string_lossy())
                         + "\""
                         + &captures[3]);
                 }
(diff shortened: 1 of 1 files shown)


davidlattimore:
Thanks @Marcono1234!

PR #272, opened 2023-02-18
Suggest offline mode when adding dependency fails

Marcono1234:
This makes it easier to discover the `:offline` command and hopefully makes it easier for users to work around issues with their internet connection.

No worries if you reject this pull request because checking for a specific error message is not a clean enough solution. Feedback is appreciated!

### DIFF
--- a/evcxr/src/cargo_metadata.rs
+++ b/evcxr/src/cargo_metadata.rs
@@ -66,12 +66,21 @@ pub(crate) fn validate_dep(dep: &str, dep_config: &str, config: &Config) -> Resu
         let primary_error_pattern = PRIMARY_ERROR_PATTERN
             .get_or_init(|| Regex::new("(.*) as a dependency of package `[^`]*`").unwrap());
         let mut message = Vec::new();
+        let mut suggest_offline_mode = false;
         for line in String::from_utf8_lossy(&output.stderr).lines() {
             if let Some(captures) = primary_error_pattern.captures(line) {
                 message.push(captures[1].to_string());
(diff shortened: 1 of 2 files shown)


davidlattimore:
Thanks @Marcono1234. Don't worry about checking for specific errors - that boat sailed long ago - although we do make slightly less use of error parsing than we used to and for the most part we're parsing errors in JSON format.

evcxr/src/cargo_metadata.rs
```diff
@@ -66,12 +66,21 @@ pub(crate) fn validate_dep(dep: &str, dep_config: &str, config: &Config) -> Resu
...
         let mut message = Vec::new();
+        let mut suggest_offline_mode = false;
         for line in String::from_utf8_lossy(&output.stderr).lines() {
             if let Some(captures) = primary_error_pattern.captures(line) {
                 message.push(captures[1].to_string());
             } else if !ignored_lines_pattern.is_match(line) {
                 message.push(line.to_owned());
             }
+
+            if line.contains("failed to fetch `https://github.com/rust-lang/crates.io-index`") {
```
kting28, line 77: Is checking the entire URL necessary?  Is `failed to fetch  ``https://` enough? asking because it's possible to repoint cargo to use crates index hosted somewhere else ...

bjorn3: Checking part could trigger on using a non-existing git dependency too. So it depends on how much false positives are acceptable.

kting28: Fair enough..  repointed crates index is an edge use case anyway (hopefully)

PR #274, opened 2023-02-19
Prompt for more input after attribute

### LINKED ISSUE #219: Attributes don’t start multiline input in evcxr_repl
**How to reproduce:**

1. Start `evcxr` binary
2. Enter `#[derive(Debug)]`

**Expected behaviour:**

REPL enters multiline mode to allow writing an item definition.

**Actual behaviour:**

REPL tries to execute this line on it’s own, resulting in compilation error.

Marcono1234:
Fixes #219

For parsing the attribute this tries to leverage the existing parsing logic of `validate_source_fragment`, e.g. for strings and brackets, and only additionally tracks whether it currently is still within the attribute or whether the item after the attribute (e.g. a struct) is expected and therefore the input should be considered incomplete.

This might not be the cleanest solution though; feedback is appreciated!

Also, please let me know in case I forget to consider cases where this erroneously detects something as attribute.

### DIFF
--- a/evcxr_repl/src/scan.rs
+++ b/evcxr_repl/src/scan.rs
@@ -91,20 +91,36 @@ pub enum FragmentValidity {
 pub fn validate_source_fragment(source: &str) -> FragmentValidity {
     use Bracket::*;
     let mut stack: Vec<Bracket> = vec![];
+    // The expected depth `stack` should have after the closing ']' of the attribute
+    // is read; None if closing ']' has already been read or currently not reading
+    // attribute
+    let mut attr_end_stack_depth: Option<usize> = None;
+    // Whether the item after an attribute is expected; is set to true after the
+    // expected attr_end_stack_depth was reached
+    let mut expects_attr_item = false;
 
     let mut input = source.char_indices().peekable();
     while let Some((i, c)) = input.next() {
+        // For simplicity set `expects_attr_item = false` and only for all cases where next
+        // char is not start of an item (e.g. comment or whitespace) restore its old value
+        let expected_attr_item = expects_attr_item;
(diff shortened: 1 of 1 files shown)


evcxr_repl/src/scan.rs
```diff
@@ -91,20 +91,36 @@ pub enum FragmentValidity {
...
+    // is read; None if closing ']' has already been read or currently not reading
+    // attribute
+    let mut attr_end_stack_depth: Option<usize> = None;
+    // Whether the item after an attribute is expected; is set to true after the
+    // expected attr_end_stack_depth was reached
+    let mut expects_attr_item = false;
 
     let mut input = source.char_indices().peekable();
     while let Some((i, c)) = input.next() {
+        // For simplicity set `expects_attr_item = false` and only for all cases where next
```
davidlattimore, line 104: I guess one alternative to this would be to have a variable like: `let mut is_attr_target = true;`. Possible alternative names `is_code`, `non_empty`. Then instead of `expects_attr_item = expected_attr_item`, you can do `is_attr_target = false`. After the match you can then update `expects_attr_item`. e.g. `if is_attr_target {expects_attr_item = false}`.

I'll leave it up to you if you think this is more/less easy to understand. I'm happy to merge either way.

davidlattimore (approved):
Thanks for the PR - this is a nice improvement!!

evcxr_repl/src/scan.rs
Marcono1234: Good idea, I am also not that happy with the `expects_attr_item = expected_attr_item` since it is a bit confusing. I have adjusted it in the latest commit, I hope that is what you had in mind.

PR #279, opened 2023-03-06
Add missing commands to COMMON.md

Marcono1234:
I have also changed the order of the commands to be alphabetical (not sure if there was previously any specific order).


davidlattimore:
Thanks!

PR #287, opened 2023-04-25
add useful resources

### LINKED ISSUE #278: Dr. Shahin Rostami's Book way to outdated
You recommend Dr. Shahin Rostami's Book "Data Analysis with Rust Notebooks" as a 3rd party resource in the evcxr_jupyter crate. I bought the book and maybe it should be added as an info that it is not up to date (2021.09). For example it still uses the plotly crate version "0.4.0" which still really lacks functionality, current version is "0.8.3".
I were a bit disappointed because in the end I had to gather all resources myself again, in which case I do not need to buy a book. :(

wiseaidev:
Just a lil contrib.

**Edit**: Closes #278


davidlattimore:
Thanks :)

PR #289, opened 2023-04-26
Add command to get variable type

d86leader:
Add commands `:type` and `:t` to get a type of a single local variable

### DIFF
--- a/evcxr/src/command_context.rs
+++ b/evcxr/src/command_context.rs
@@ -368,6 +368,20 @@ Panic detected. Here's some useful information if you're filing a bug report.
                 },
             ),
             AvailableCommand::new(
+                ":type",
+                "Show variable type",
+                |ctx, _state, args| {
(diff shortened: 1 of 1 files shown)


evcxr/src/command_context.rs
```diff
@@ -586,6 +600,29 @@ Panic detected. Here's some useful information if you're filing a bug report.
...
+            bail!("Variable name required")
+        };
+
+        let mut out = None;
+        for (var, ty) in self.eval_context.variables_and_types() {
+            if var == args {
+                out = Some(ty.to_owned());
+                break;
+            }
+        }
```
d86leader, line 617: Alternatively one could add a public function to eval_context to lookup a single variable. Not sure what is better, so I went with less intrusive changes and did like in vars_as_text.

davidlattimore:
Thanks for the PR!

PR #290, opened 2023-04-28
Make :dep print info about compiled crates

### LINKED ISSUE #224: Any verbosity option for `:dep`?

d86leader:
Closes #224 Obsoletes #267

When cargo compiles crates, the name of the crate is printed to stderr. Example output:

```
Welcome to evcxr. For help, type :help
>> :dep regex
   Compiling memchr v2.5.0
   Compiling regex v1.8.1
>>
```

### DIFF
--- a/evcxr/src/module.rs
+++ b/evcxr/src/module.rs
@@ -288,13 +288,41 @@ offline = {}
         )
     }
 }
+
+/// Run a cargo command prepared for the provided `code_block`, processing the
+/// command's output.
 fn run_cargo(
     mut command: std::process::Command,
     code_block: &CodeBlock,
 ) -> Result<std::process::Output, Error> {
-    let cargo_output = match command.output() {
+    let mb_child = command
+        .stderr(std::process::Stdio::piped())
+        .stdout(std::process::Stdio::piped())
+        .spawn();
+    let mut child = match mb_child {
         Ok(out) => out,
         Err(err) => bail!("Error running 'cargo rustc': {}", err),
+    };
+
+    let mut stdout = child.stdout.take().unwrap();
+    let stderr = std::io::BufReader::new(child.stderr.take().unwrap());
+    let mut all_output = Vec::new();
+    let mut all_errors = Vec::new();
+
+    use std::io::{BufRead, Read};
+    for mb_line in stderr.split(10) {
+        let mut line = mb_line?;
+        tee_error_line(&line);
+        all_errors.append(&mut line);
+    }
+    stdout.read_to_end(&mut all_output)?;
+
+    let status = child.wait()?;
+
+    let cargo_output = std::process::Output {
+        status,
+        stdout: all_output,
+        stderr: all_errors,
     };
     if cargo_output.status.success() {
         Ok(cargo_output)
@@ -313,6 +341,26 @@ fn run_cargo(
             }
         } else {
             bail!(Error::CompilationErrors(errors));
+        }
+    }
+}
+
+/// Process one line from cargo, either copying it to stderr or ignoring.
+///
+/// At this point it looks for messages about compiling dependency crates.
+fn tee_error_line(line: &[u8]) {
+    use std::io::Write;
+    static CRATE_COMPILING: OnceCell<regex::bytes::Regex> = OnceCell::new();
+    let crate_compiling = CRATE_COMPILING
+        .get_or_init(|| regex::bytes::Regex::new("^\\s*Compiling (\\w+)(?:\\s+.*)?$").unwrap());
+    if let Some(captures) = crate_compiling.captures(line) {
+        let crate_name = captures.get(1).unwrap().as_bytes();
+        if crate_name != CRATE_NAME.as_bytes() {
+            // write line and the following nl symbol as it was stripped before
+            std::io::stderr()
+                .write_all(line)
+                .expect("Writing to stderr should not fail");
+            eprintln!();
         }
     }
 }
--- a/evcxr/src/module.rs
+++ b/evcxr/src/module.rs
@@ -295,6 +295,8 @@ fn run_cargo(
     mut command: std::process::Command,
     code_block: &CodeBlock,
 ) -> Result<std::process::Output, Error> {
+    use std::io::{BufRead, Read};
+
     let mb_child = command
         .stderr(std::process::Stdio::piped())
         .stdout(std::process::Stdio::piped())
@@ -304,20 +306,26 @@ fn run_cargo(
         Err(err) => bail!("Error running 'cargo rustc': {}", err),
     };
 
+    // Collect stdout in a parallel thread
     let mut stdout = child.stdout.take().unwrap();
-    let stderr = std::io::BufReader::new(child.stderr.take().unwrap());
-    let mut all_output = Vec::new();
-    let mut all_errors = Vec::new();
+    let output_thread = std::thread::spawn(move || {
+        let mut buf = Vec::new();
+        stdout.read_to_end(&mut buf)?;
+        Ok::<_, Error>(buf)
+    });
 
-    use std::io::{BufRead, Read};
+    // Collect stderr synchronously
+    let stderr = std::io::BufReader::new(child.stderr.take().unwrap());
+    let mut all_errors = Vec::new();
     for mb_line in stderr.split(10) {
         let mut line = mb_line?;
         tee_error_line(&line);
         all_errors.append(&mut line);
+        all_errors.push(10);
     }
-    stdout.read_to_end(&mut all_output)?;
 
     let status = child.wait()?;
+    let all_output = output_thread.join().expect("Panic in child thread")?;
 
     let cargo_output = std::process::Output {
         status,
--- a/evcxr/src/command_context.rs
+++ b/evcxr/src/command_context.rs
@@ -620,7 +620,7 @@ Panic detected. Here's some useful information if you're filing a bug report.
             let out = format!("{args}: {out}");
             Ok(EvalOutputs::text_html(out.clone(), out))
         } else {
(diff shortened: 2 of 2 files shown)


davidlattimore (commented):
Thanks for picking this up!

evcxr/src/module.rs
```diff
@@ -288,14 +288,42 @@ offline = {}
...
         Err(err) => bail!("Error running 'cargo rustc': {}", err),
     };
+
+    let mut stdout = child.stdout.take().unwrap();
+    let stderr = std::io::BufReader::new(child.stderr.take().unwrap());
+    let mut all_output = Vec::new();
+    let mut all_errors = Vec::new();
+
+    use std::io::{BufRead, Read};
+    for mb_line in stderr.split(10) {
```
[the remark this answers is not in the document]
bjorn3: In UTF-8 all multi-byte characters have the high bit set in every byte.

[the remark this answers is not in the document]
davidlattimore: Ah, you're right, thanks! Nevermind then :)

evcxr/src/module.rs
```diff
@@ -288,14 +288,42 @@ offline = {}
...
+        let mut line = mb_line?;
+        tee_error_line(&line);
+        all_errors.append(&mut line);
+    }
+    stdout.read_to_end(&mut all_output)?;
```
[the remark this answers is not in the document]
sigmaSd: Nice how you preemptively figured this, I remember getting this bug , and spending some good amount of time debugging it , here is my current code 
```rs
use std::{
    io::Read,
    process::{Child, Output},
    sync::mpsc,
};
pub trait ProcessUtils {
    fn interactive_output(self, function: Option<fn(&mut Child) -> Result<()>>) -> Result<Output>;
}

impl ProcessUtils for std::processs::Child {
    fn interactive_output(
        mut self,
        function: Option<fn(&mut Child) -> Result<()>>,
    ) -> Result<Output> {
        let mut stdout = self.stdout.take().expect("stdout is piped");
        let mut stderr = self.stderr.take().expect("stderr is piped");

        let (tx_out, rx) = mpsc::channel();
        let tx_err = tx_out.clone();
        enum OutType {
            Stdout(Vec<u8>),
            Stderr(Vec<u8>),
        }

        std::thread::spawn(move || {
            let mut out = Vec::new();
            let _ = stdout.read_to_end(&mut out);
            let _ = tx_out.send(OutType::Stdout(out));
        });

        std::thread::spawn(move || {
            let mut err = Vec::new();
            let _ = stderr.read_to_end(&mut err);
            let _ = tx_err.send(OutType::Stderr(err));
        });

        while self.try_wait()?.is_none() {
            if let Some(ref function) = function {
                function(&mut self)?;
            }
        }
        let mut stdout = None;
        let mut stderr = None;
        for _ in 0..2 {
            match rx.recv()? {
                OutType::Stdout(out) => stdout = Some(out),
                OutType::Stderr(err) => stderr = Some(err),
            }
        }

        Ok(Output {
            status: self.wait()?,
            stdout: stdout.unwrap(),
            stderr: stderr.unwrap(),
        })
    }
}
```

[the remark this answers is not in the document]
d86leader: My thinking here was that cargo usually writes nothing or very little to stdout; the build scripts can write something, but rarely do. My thought thus was that the buffer would only fill up in rare cases. I'm not going to say no to being more secure though.

Would like to go without threads here, but we don't have mio and can't rely on epoll, so sure.

MPSC is a bit of an overkill for the goal of writing once, but nice approach.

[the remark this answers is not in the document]
davidlattimore: You probably don't need MPSC. The thread can construct a `Vec<u8>` and then just return it. When you join the thread, you get the `Vec`.

evcxr/src/module.rs
```diff
@@ -288,14 +288,42 @@ offline = {}
...
+    let mut stdout = child.stdout.take().unwrap();
+    let stderr = std::io::BufReader::new(child.stderr.take().unwrap());
+    let mut all_output = Vec::new();
+    let mut all_errors = Vec::new();
+
+    use std::io::{BufRead, Read};
+    for mb_line in stderr.split(10) {
+        let mut line = mb_line?;
+        tee_error_line(&line);
+        all_errors.append(&mut line);
```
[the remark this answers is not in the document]
d86leader: Nice catch, fixed now

evcxr/src/module.rs
```diff
@@ -317,6 +353,26 @@ fn run_cargo(
...
 }
 
+/// Process one line from cargo, either copying it to stderr or ignoring.
+///
+/// At this point it looks for messages about compiling dependency crates.
+fn tee_error_line(line: &[u8]) {
+    use std::io::Write;
+    static CRATE_COMPILING: OnceCell<regex::bytes::Regex> = OnceCell::new();
+    let crate_compiling = CRATE_COMPILING
+        .get_or_init(|| regex::bytes::Regex::new("^\\s*Compiling (\\w+)(?:\\s+.*)?$").unwrap());
```
Marcono1234, line 363: @d86leader, it looks like this can fail at runtime. When I tried to install evcxr from GitHub, shortly after starting, it panicked:
```
Welcome to evcxr. For help, type :help
>> thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: Syntax(
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
regex parse error:
    ^\s*Compiling (\w+)(?:\s+.*)?$
     ^^
error: Unicode-aware Perl class not found (make sure the unicode-perl feature is enabled)
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
)', evcxr\src\module.rs:363:87
```

Might be necessary to enable the [`unicode-perl` feature](https://docs.rs/regex/1.8.1/regex/#unicode-features) for the `regex` dependency. Or adjust the pattern to not use any of these character classes but instead only something like `[ \t]`, ... (and maybe add a comment then above the pattern warning that no Unicode character classes should be used)

d86leader: Interesting, it's surprising to me that the behaviour is not reliable between systems. What OS do you run it on, and do you do some unusual compilation? I'll try to reproduce and enable the required features.

Marcono1234: I am using Windows 10, with Cargo 1.68.2 (6feb7c9cf 2023-03-26)

Also, I initially started building with `cargo install evcxr_repl`, then cancelled this and ran `cargo install --force --git https://github.com/evcxr/evcxr.git evcxr_repl`. Not sure if that had any negative effect, but I don't think it should.

davidlattimore: Thanks for noticing this! I think the reason it's not happening when building from a source checkout is because of the cargo workspace. `evcxr_jupyter` depends on `zeromq`, which depends on `regex` with default features enabled.

d86leader: Ok, makes sense. This patch fixes the problem:
```
diff --git a/evcxr/src/module.rs b/evcxr/src/module.rs
index 2de1105..6e4dbec 100644
--- a/evcxr/src/module.rs
+++ b/evcxr/src/module.rs
@@ -359,8 +359,9 @@ fn run_cargo(
 fn tee_error_line(line: &[u8]) {
     use std::io::Write;
     static CRATE_COMPILING: OnceCell<regex::bytes::Regex> = OnceCell::new();
-    let crate_compiling = CRATE_COMPILING
-        .get_or_init(|| regex::bytes::Regex::new("^\\s*Compiling (\\w+)(?:\\s+.*)?$").unwrap());
+    let crate_compiling = CRATE_COMPILING.get_or_init(|| {
+        regex::bytes::Regex::new("^[ \t]*Compiling ([a-zA-Z0-9_-]+)(?:[ \t]+.*)?$").unwrap()
+    });
     if let Some(captures) = crate_compiling.captures(line) {
         let crate_name = captures.get(1).unwrap().as_bytes();
         if crate_name != CRATE_NAME.as_bytes() {
```
I can send a new pull request if that's easier for you

Marcono1234: > This patch fixes the problem

@d86leader, might be good to add a comment above the regex though explaining why the character classes (e.g. `\w`) are not used. Otherwise without comment the next person who comes across this code might think the regex can be simplified, and reintroduces this issue.

d86leader: Would also be good to add a build with the specified regex version. That way noone would also be able to create new regexes with the same problem.

Marcono1234: @d86leader, just to make sure there is no misunderstanding, do you want to submit a pull request (either adjusting the regex pattern or enabling the `unicode-perl` feature), or should I do that?

davidlattimore: Sorry, I mean to submit the patch above, but forgot, although I'm wondering if we should just enable `unicode-perl`. It doesn't appear to slow down compilation of the regex crate - not compared with enabling all regex features anyway. Happy to receive a PR from anyone for either approach - otherwise I'll make the change myself.

Marcono1234: If this is only a small change (and if you have tested it already anyway), maybe it would be easiest if you do it yourself, so that you can choose the solution you prefer.

davidlattimore: Thanks for the prompt. I've pushed a fix

Issue #297, opened 2023-05-24
Allow extra command line arguments


FlashSheridan:
Rust newbies (trust me) sometimes need to see what gets passed in as command-line arguments, via env::args or more complicated techniques.  Currently std::env::args() just has something like `Args { inner: ["~/.cargo/bin/evcxr"] }`, even if `--opt 2` was passed, and extra arguments stop evcxr from launching.  Currently I need to pass in and inspect a bunch of arguments escaped in different ways, so I’d like evcxr just to accept them and put them in std::env::args(), perhaps after an `--extra-args` argument.

davidlattimore:
Evcxr has two processes, both of which are the `evcxr` binary. The parent process is the one that you run. It's what displays the prompt, parses what you type, writes source files to disk, runs `cargo` etc. Then there's a subprocess, which is where your code get executed. When you run `std::env::args`, it's looking at the arguments passed to this subprocess.

So I think it would be possible to pass extra arguments to this subprocess, which would, by default, ignore them. If you clone the evcxr repository, then edit `evcxr/src/child_process.rs` and make a change like the following:

```patch
diff --git a/evcxr/src/child_process.rs b/evcxr/src/child_process.rs
index ef15635..1bdaf90 100644
--- a/evcxr/src/child_process.rs
+++ b/evcxr/src/child_process.rs
@@ -37,6 +37,7 @@ impl ChildProcess {
             bail!("Our current binary doesn't call runtime_hook()");
         }
         command
+            .arg("--foo")
             .env(runtime::EVCXR_IS_RUNTIME_VAR, "1")
             .env("RUST_BACKTRACE", "1")
             .stdin(std::process::Stdio::piped())
```

If you then run `cargo run -p evcxr_repl` and in the resulting shell do `std::env::args()`, you get:

```
Args { inner: ["/home/david/work/evcxr/target/debug/evcxr", "--foo"] }
```

Hopefully that helps

PR #298, opened 2023-05-25
Allow extra command line arguments #297

FlashSheridan:
- Pass, per the maintainer’s suggestion, extra arguments to the child subprocess, so that they can be seen in the REPL.
- Add, for Rust newbies like me, a new command-line option to allow more command-line arguments:

```--extra-args <extra-args>...    Extra arguments; ignored, but show up in std::env::args()```

I am, by hypothesis, a Rust newbie, so review may be in order.  

## Testing

- `EVCXR_TMPDIR=$HOME/tmp/e1 cargo test -- --test-threads 1` passed on MacOS 12.6.2 21G320 on M1 Max, with the same two crash dialogs I got from main. 
- `target/debug/evcxr --opt 2 --edit-mode emacs --extra-args \\--foo bar baz` gave: 

`>> std::env::args()`

`Args { inner: ["/Users/flash/Documents/Code/external_code/evcxr/target/debug/evcxr", "target/debug/evcxr", "--opt", "2", "--edit-mode", "emacs", "--extra-args", "\\--foo", "bar", "baz"] }`

### DIFF
--- a/evcxr/src/child_process.rs
+++ b/evcxr/src/child_process.rs
@@ -37,6 +37,7 @@ impl ChildProcess {
             bail!("Our current binary doesn't call runtime_hook()");
         }
         command
+            .arg("--foo")
             .env(runtime::EVCXR_IS_RUNTIME_VAR, "1")
             .env("RUST_BACKTRACE", "1")
             .stdin(std::process::Stdio::piped())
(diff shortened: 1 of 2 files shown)


davidlattimore:
Thanks!

PR #309, opened 2023-07-18
Add instructions for installation inside virtualenv

nils-werner:
Many people use jupyter inside a virtualenv. In this case, the variable `JUPYTER_PATH` should be used to provide the correct installation directory.


davidlattimore:
Sorry, I really dropped the ball on this. Thanks for the contribution!

PR #317, opened 2023-08-23
add a note on how to find the config dir

### LINKED ISSUE #314: `init.evcxr` and `prelude.rs` placed in the suggested MacOS location doesn't seem to do anything
Hello, and thank you for the amazing crate.

I'm confused about the directions in COMMON.md, which recommends mac users create a file `/Users/anand/Library/Preferences/evcxr/init.evcxr` and `prelude.rs`

After placing :sccache 1 and a local dependence in the init file and a simple variable assignment in prelude, it doesn't seem that it gets looked at or loaded. I don't see any message like `Executing prelude from "~/.config/evcxr/prelude.rs"` as shown in COMMON.md. 

Is there something else I missed?

anandijain:
I didn't modify the table, since I'm not sure if it is always in `/Users/anand/Library/Application Support/evcxr/init.evcxr`. 

This should help to resolve #314


davidlattimore:
Good idea, thanks!

PR #323, opened 2023-09-11
Adding a command `:doc` to show docmentation.

baiguoname:
Adding a command `:doc` to show docmentation of a function, module, keyword, variable etc...
For eaxample: `:doc i32`, `:doc std`, `:doc std::fs::File::create`.
In jupter, it would be better if the documentation can be shown in a style of markdown.


evcxr/src/eval_context.rs
```diff
@@ -508,6 +508,31 @@ impl EvalContext {
...
+        let (modified_code, hover_offset) = if code == "let" {
+            (String::from("let _ = 1;"), 0)
+        } else {
+            (format!("{};", code), code.len() as u32)
+        };
+        let (user_code, code_info) = CodeBlock::from_original_user_code(&modified_code);
+        let user_code = state.apply(user_code, &code_info.nodes)?;
+        let pad_code = state.analysis_code(user_code);
+        self.analyzer.set_source(pad_code.code_string())?;
+        //if I set the hover position to be the start position of input code, then
```
davidlattimore, line 525: That makes sense and is consistent with what I observe if I hover over the "std" in vscode

evcxr/src/eval_context.rs
```diff
@@ -508,6 +508,31 @@ impl EvalContext {
...
+        } else {
+            (format!("{};", code), code.len() as u32)
+        };
+        let (user_code, code_info) = CodeBlock::from_original_user_code(&modified_code);
+        let user_code = state.apply(user_code, &code_info.nodes)?;
+        let pad_code = state.analysis_code(user_code);
+        self.analyzer.set_source(pad_code.code_string())?;
+        //if I set the hover position to be the start position of input code, then 
+        //`:doc std::fs::File::create` will return the information of `std`, rather than `create`
+        let wrapped_offset = pad_code.user_offset_to_output_offset_for_hover()? as u32 + hover_offset;
```
davidlattimore, line 527: I tried changing this to the following and it worked for me (including in jupyter):

```rust
let wrapped_offset = pad_code.user_offset_to_output_offset(hover_offset)? as u32;
```

I guess if you're passing `hover_offset` as `code.len()` then you'll be getting the hover for the `;`, so you could consider subtracting 1 to get the last character of what the user typed, but even without subtracting 1 it still seemed to work for me.

evcxr/src/command_context.rs
```diff
@@ -619,6 +626,16 @@ Panic detected. Here's some useful information if you're filing a bug report.
...
     }
+
+    fn hover(&mut self, state: &mut ContextState, args: &Option<String>) -> Result<EvalOutputs, Error> {
+        let args = if let Some(x) = args {
+            x.trim()
+        } else {
+            bail!("Input required")
+        };
+        let res = self.eval_context.hover(args, state)?;
+        Ok(EvalOutputs::text_html(res.clone(), res))
```
davidlattimore, line 637: The first argument to `text_html` should ideally be plain text and the second argument HTML. At the moment you're passing markdown for both.

It looks like the two formats that rust-analyzer supports are plain text and markdown. For the first argument, it could be good to pass actual plain text. For the second argument, one possibility would be to use the `markdown` crate to convert the markdown to HTML.

davidlattimore (commented):
Nice work! This is great!

evcxr/src/eval_context.rs
baiguoname: I test with 
```
let wrapped_offset = pad_code.user_offset_to_output_offset(hover_offset)? as u32;
```
It works fine if I don't include `init.evcxr` and `prelude.rs`. But after I include `init.evcxr` and `prelude.rs`, then I have the problem before, it seems `wrapped_offset` didn't get the right position. This is the sorce code when I include `init.evcxr` and `prelude.rs`:
```
#![allow(unused_imports, unused_mut, dead_code)]
#![allow(unused_imports, unused_variables, non_camel_case_types)]

use serde::Serialize;
use mongodb::Client;
use serde::Deserialize;
use chrono::Timelike;
use chrono::Duration;
use bt::prelude::*;
use bt::std_prelude::*;
use pytest::prelude::*;
use stra::prelude::*;
use getdata::prelude::*;

            struct EvcxrUserCodeError {}
            impl<T: std::fmt::Display> From<T> for EvcxrUserCodeError {
                fn from(error: T) -> Self {
                    eprintln!("{}", error);
                    
                    EvcxrUserCodeError {}
                }
            }
        
fn evcxr_variable_store<T: 'static>(_: T) {}
#[allow(unused_variables)]
async fn evcxr_analysis_wrapper(
) -> Result<(), EvcxrUserCodeError> {
i32;
Ok(())
}
```

evcxr/src/command_context.rs
baiguoname: I fixed it in this way: 
I get both the plain text and markdown, and then convert the markdown with using a simple markdown html (I am trying to use a more proper markdown crate mdbook(https://github.com/rust-lang/mdBook)). And then parse the two to method `text_html`.

evcxr/src/eval_context.rs
davidlattimore: Turned out to be an issue with attributes not being properly committed. I've pushed a [fix](https://github.com/evcxr/evcxr/commit/eca319a1f54f7408c03d11640774416528cc4624) it.

A more detailed explanation in case you're interested:

`CodeBlock::user_offset_to_output_offset` relies on the code that the user is currently executing having kind `CodeKind::OriginalUserCode` and any code that the user previously executed having a different kind. The method `EvalContext::commit_old_user_code` commits already-executed user code by changing its origin from `CodeKind::OriginalUserCode` to `CodeKind::OtherUserCode`, however it missed doing so for attributes (like `#![allow(...)]`), which were a relatively recent addition to evcxr. This meant that `CodeBlock::user_offset_to_output_offset` thought that the attribute was part of what was being executed now, so for example offset 0 would be the `#`.

baiguoname:
Now, it works fine in Jupyter:
image

davidlattimore:
Looks like the new test needs updating (text/plain vs text/html).

Can you also run `cargo fmt` and `cargo clippy` (fixing any warnings that clippy reports)?

After that, I think we should be good to merge.

baiguoname:
I run the `examples/example_hover.rs` well in local, but it seems fail to check, maybe I should delete this example?

davidlattimore:
> I run the `examples/example_hover.rs` well in local, but it seems fail to check, maybe I should delete this example?

I think it's complaining that it isn't formatted, but sure, deleting it probably makes sense, since I guess it was mostly useful for manual testing during development.

PR #325, opened 2023-09-13
Make the error less mysterious when `rust_analyzer` can't determin the variable type.

baiguoname:
When the input is like:
```
let a = undefined_function(1);
```
the error show needing of type anotation, but it is more direct if the error show that then function `undefined_function` is not defined.
I add code to call `check` after `fix_variable_types`, it works fine in crate `evcxr`, but in jupyter, if the input is the above example and include `evcxr.init` and `prelude.rs`(as like https://github.com/evcxr/evcxr/pull/323#discussion_r1322282512), the cell will run forever(never teminate). If I remove `evcxr.init` and `prelude.rs`, then it works fine. I thought it may be a problem related to `commit_old_user_code`, but I can't figure it out.

### DIFF
--- a/evcxr/src/eval_context.rs
+++ b/evcxr/src/eval_context.rs
@@ -441,7 +441,7 @@ impl EvalContext {
         let mut phases = PhaseDetailsBuilder::new();
         let code_out = state.apply(user_code.clone(), &code_info.nodes)?;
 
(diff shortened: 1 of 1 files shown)


davidlattimore:
Thanks for the fix!

PR #327, opened 2023-09-16
Adding a command `::bench` to benchmark an input function.

baiguoname:
In jupyter, when user input the command `::bench` with following a function define body in the cell, there is a project named `bench` will be generated in the path `context.config.crate_dir()`. The project has a dependency on crate `criterion`, and the `lib.rs` contains the user input function. Then command  `cargo bench` will be run in the project path. After that, the benchmark report will be generated, which then is sent to jupyter output.
There is an example:
图片


evcxr/Cargo.toml
```diff
@@ -52,6 +52,7 @@ ariadne = "0.3.0"
 triomphe = "0.1.8"
 
 pulldown-cmark = { version = "0.9.3", default-features = false, features = ["simd"] }
+tl = "0.7"
```
bjorn3, line 55: What is the reason to choose this over something like html5ever. It shouldn't matter if a slower html parser is used, right?

baiguoname: I didn't choose `tl`, I just know it from a friend yesterday. I not fimalier of `.html` , is there an obvious perfomance gap between the two crates?

bjorn3: `tl` claims it is about 10x faster at the cost of not entirely following the html spec. `html5ever` is a lot more popular and does follow the html spec. For this use case I don't think the performance difference should matter, so sticking with `html5ever` may be a better option.

baiguoname: Thanks for your advice, I will change the crate dependency and make a new commit.
There is an another difficulty I am struggling at:
The generated `index.html` from report directory has some `<href = some_svg.svg>` hyperlink, which could be clicked at the the showing `.html` and jump to source `some_svg.svg` file. Is there a way I can put the all  these `.svg` files into `index.html`, and then send the `index.html` to jupyter, which the showing shape of `index.html` will be the same as original one?

bjorn3: You could encode the `<a href="some_svg.svg">` as `<a href="data:image/svg+xml,base64;AAAA=" target="_blank">` where `AAAA=` is the base64 encoded contents of `some_svg.svg`. I'm not sure if the browser will attempt to show it or download it though. If it attempts to download it you may need to use javascript if jupyter notebook doesn't block that.

davidlattimore:
Interesting feature!

Having a separate "bench" directory with separate `Cargo.toml` has some advantages and disadvantages. An advantage is that it perhaps makes this change affect less of the existing evcxr code. It also means that subsequent non-bench executions won't be affected. However unless I'm mistaken it also means that if the function that you're benchmarking requires any 3rd party crates that the user already added with `:dep`, it won't work. Similarly, any existing functions or types defined by the user, won't be available. Also features like tab-completion and inline errors won't work in a benchmark block. I'm unsure what the best way to integrate is, but at the moment it feels like a very separate feature to the rest of the jupyter kernel in ways that might be surprising to a user.

Is it necessary to have the command start with a double-colon, or would it be possible to make it start with a single colon like the existing commands?

baiguoname:
> Interesting feature!
> 
> Having a separate "bench" directory with separate `Cargo.toml` has some advantages and disadvantages. An advantage is that it perhaps makes this change affect less of the existing evcxr code. It also means that subsequent non-bench executions won't be affected. However unless I'm mistaken it also means that if the function that you're benchmarking requires any 3rd party crates that the user already added with `:dep`, it won't work. Similarly, any existing functions or types defined by the user, won't be available. Also features like tab-completion and inline errors won't work in a benchmark block. I'm unsure what the best way to integrate is, but at the moment it feels like a very separate feature to the rest of the jupyter kernel in ways that might be surprising to a user.
> 
> Is it necessary to have the command start with a double-colon, or would it be possible to make it start with a single colon like the existing commands?

The Double-colon is mimicking on `ipython` magic command. `ipython` has two types of magic command: line command like `line command` `%time` and `cell command` `%%time`. The former functions on a line, and the latter functions on a jupyter cell. And another reason is that I use Double-colon so that I don't have to inject code in existed function like `from_orginal_user_code` from `code_blocks.rs`, then I can remove the adding code easily if it is needed.
Adding this command do inject many code and adding more dependencies, and since you feel this command to be a separate feature to the rest of jupyter kernel, I will withdraw this pr for more detail consideration.

PR #330, opened 2023-09-24
Ignore potential poisoning of tokio runtime mutex

### LINKED ISSUE #329: Cached tokio runtime after panic
System: windows
evcxr: 0.15.1

With the following init block

```
:clear
:dep tokio = {version = "1.17.0", features = ["full"]}
```
executing the following cell twice leads to two different behaviours:

```rust
async fn foo() {
    panic!("hello")
}

foo().await
```

First execution:

> thread '<unnamed>' panicked at 'hello', src/lib.rs:3:5
stack backtrace:
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.`

Second execution:

> thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: PoisonError { .. }', src/lib.rs:100:63
stack backtrace:
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.

Obviously, there is nothing one can do about the first error, however, the second error is problematic, as the only way recovering from this seems to be to issue a `:clear` and `dep tokio` again. This is "bad", because all other state is also lost.

I assume this `PoisonError` stems from the mutex on the runtime introduced in #90.

It would be nice if we could either: 
 - only clear the cached runtime
 - or in case of a panic in the `block_on` thread automatically invalidate the cached runtime to create a new one on the next async call.

This should then allow to keep using variables that eventually were created in a different runtime (as long as they are Sync/Send...)??!

Either way, like what you are doing here - if I find some time and you are interested in contributions I might actually give it a shot ;-)

martinitus:
Resolves #329

Not sure if this is the best way to go - maybe droping the entire `Arc<Mutex<Runtime>>` and creating a fresh one is better.

However, below seems to work for me :-)

### DIFF
--- a/evcxr/src/eval_context.rs
+++ b/evcxr/src/eval_context.rs
@@ -1531,12 +1531,16 @@ impl ContextState {
         }
         if self.async_mode {
             user_code = CodeBlock::new()
-                .generated(stringify!(evcxr_variable_store
-                    .lazy_arc("evcxr_tokio_runtime", || std::sync::Mutex::new(
-                        tokio::runtime::Runtime::new().unwrap()
-                    ))
-                    .lock()
-                    .unwrap()))
+                .generated(stringify!(
+                let mut mutex = evcxr_variable_store.lazy_arc("evcxr_tokio_runtime",
+                    || std::sync::Mutex::new(tokio::runtime::Runtime::new().unwrap())
+                );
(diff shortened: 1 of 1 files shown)


davidlattimore:
Thanks for the PR. Happy to go with this and we can always revisit if it turns out that creating a new runtime is better for some reason.

Issue #341, opened 2023-11-05
RFC: Adding Support for Shell Command Execution


wiseaidev:
## Abstract
This RFC proposes an enhancement to the Rust Jupyter Notebook project. Specifically, it suggests adding the capability to execute shell commands directly within Rust notebook cells, similar to the way Python kernels allow the execution of shell commands with a leading exclamation mark (!).

## Motivation
The motivation behind this proposal stems from the desire to simplify the experience of Rust developers using Jupyter notebooks. In Python kernels, developers can effortlessly execute Bash commands by prefixing them with an exclamation mark, such as `!ls`. Unfortunately, this feature is not available in Rust cells.

## Proposal
I propose the addition of a feature that permits users to execute shell commands from within Rust notebook cells by prefixing the command with an exclamation mark (!), or any special character. Much like the existing behavior in Python kernels, this would enable users to perform common shell tasks such as listing files or displaying file contents with the convenience of a familiar syntax.

### Example Usage
Here is an example of how this feature would work within a Rust notebook cell:

```rust
!ls
!cat file_name
```

In the code above, the exclamation mark (!) signals that the subsequent text should be treated as a shell command, allowing the user to list files or view the content of a file directly within the Rust notebook.

## Workaround
As of now, if you want to execute Bash commands within a Rust Jupyter notebook cell, you need to employ a workaround using Rust's standard library. You can achieve this by utilizing the following code snippet:

```rust
use std::process::{Command, Output, Stdio};

// A helper function to execute a shell command from a Rust script
fn execute_command(command: &str) -> Result<(), std::io::Error> {
    let status = Command::new("bash")
        .arg("-c")
        .arg(command)
        .stderr(Stdio::inherit())
        .status()?;

    if status.success() {
        Ok(())
    } else {
        Err(std::io::Error::from_raw_os_error(status.code().unwrap_or(1)))
    }
}
```

[This workaround is functional](https://github.com/wiseaidev/rust-lang-book/blob/main/chapter-1/chapter-1.ipynb), but it lacks the simplicity and user-friendliness that we desire in Jupyter notebooks.

## Conclusion
The addition of shell command execution within Rust Jupyter Notebook cells, denoted by the exclamation mark (!), offers a valuable enhancement to the Rust programming experience. This feature would improve the interactivity and versatility of the notebook environment, making it even more accessible and powerful for the Rust community.

Love,
Mahmoud

davidlattimore:
[...]

PR #343, opened 2023-11-12
impl sh cmd

wiseaidev:
Related to #341

### DIFF
--- a/evcxr/src/command_context.rs
+++ b/evcxr/src/command_context.rs
@@ -9,6 +9,7 @@ use crate::code_block::CodeBlock;
 use crate::code_block::CodeKind;
 use crate::code_block::CommandCall;
 use crate::code_block::Segment;
+use crate::code_block::ShellCommand;
 use crate::code_block::{self};
 use crate::crash_guard::CrashGuard;
 use crate::errors::bail;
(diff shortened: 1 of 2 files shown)


evcxr/src/command_context.rs
```diff
@@ -159,6 +166,55 @@ Panic detected. Here's some useful information if you're filing a bug report.
...
+            Ok(output) => {
+                let stdout_str = String::from_utf8_lossy(&output.stdout);
+                let stderr_str = String::from_utf8_lossy(&output.stderr);
+
+                let mut eval_outputs = EvalOutputs::default();
+
+                // Add the stdout and stderr to eval_outputs
+                eval_outputs
+                    .content_by_mime_type
+                    .insert("stdout".to_string(), stdout_str.into());
```
davidlattimore, line 190: The keys in this map are mime types - i.e. "text/plain" rather than "stdout". Perhaps concatenate stdout and stderr into the one output. e.g.

```rust
eval_outputs.content_by_mime_type.insert(
                    "text/plain".to_string(),
                    stdout_str.into_owned() + stderr_str.as_ref(),
                );
```

The way it was, I didn't see any output when testing, but once changed to emit as "text/plain", the output showed up. I tested from the REPL and using jupyter notebook from a browser. I'm assuming it worked for you as it was - perhaps you were testing from a different jupyter frontend that maybe displays content with unknown mime types.

evcxr/src/command_context.rs
```diff
@@ -159,6 +166,55 @@ Panic detected. Here's some useful information if you're filing a bug report.
...
+                    "Failed to execute shell command: {}",
+                    error
+                )))
+            }
+        }
+    }
+
     pub fn set_opt_level(&mut self, level: &str) -> Result<(), Error> {
         self.eval_context.set_opt_level(level)
     }
```
davidlattimore, line 220: It can't see any way to add a comment on the appropriate line in Github, but could you also filter out shell commands in prepare_for_analysis below? At the moment, if I have a command followed by some Rust code, then I get errors reported (red squiggles) in Jupyter Notebook. This appears to be because the shell command is ending up in the code being passed to cargo check.

image

davidlattimore (commented):
Thanks for doing this, it's great!

evcxr/src/command_context.rs
bjorn3: Rather than concatenating stdout and stderr, I think a single pipe or file should be used for both stdout and stderr to preserve the ordering of both just like it would in a terminal.

wiseaidev:
Hey @davidlattimore, sorry about the delay. i lost track of this PR. Feel free to coauthor this PR if you wish.

davidlattimore:
This got picked up and merged in #395. Thanks for doing most of the work on this!

PR #347, opened 2023-11-21
Add option to set the initializing config from the current working directory.

baiguoname:
There is a discussion in [there](https://github.com/evcxr/evcxr/discussions/345).
This pr add an option to set the initializing config from the current diretory where the notebook is ready to lanuch. The config file looks like this:
```rust
[tmpdir]
tmpdir

[init]
:dep chrono = "*"

[prelude]
use chrono::NaiveDateTime as dt;
```
The attribute `tmpdir` points to the directory where the code will be compiled into, and the attribute `init` is the content that should be in  `init.evcxr`  before, and the attribute `prelude` is the content that should be in `prelude.rs` before.
This configuration file should be located in the current directory, and be named `init.evcxr`. When  `evcxr` is starts, it will check if the current directory contains the file `init.evcxr`. If not, next it will check if `crate::config_dir` contains the file `init.evcxr`. If not, it will load not anything the conmmand `:load_config` is executed. 
After loading the configuration, if it contains the attribute `tmpdir`, `evcxr` will use this `tmpdir` as  temprory compiling directory. If not, `evcxr` will check if there is an environmental variale `EVCXR_TMPDIR`.

### DIFF
--- a/evcxr/src/command_context.rs
+++ b/evcxr/src/command_context.rs
@@ -18,6 +18,7 @@ use crate::errors::Span;
 use crate::errors::SpannedMessage;
 use crate::eval_context::ContextState;
 use crate::eval_context::EvalCallbacks;
+use crate::eval_context::InitConfig;
 use crate::rust_analyzer::Completion;
 use crate::rust_analyzer::Completions;
 use crate::EvalContext;
@@ -225,27 +226,14 @@ Panic detected. Here's some useful information if you're filing a bug report.
         Ok(completions)
     }
 
-    fn load_config(&mut self, quiet: bool) -> Result<EvalOutputs, Error> {
+    fn load_config(&mut self, _quiet: bool) -> Result<EvalOutputs, Error> {
         let mut outputs = EvalOutputs::new();
-        if let Some(config_dir) = crate::config_dir() {
-            let config_file = config_dir.join("init.evcxr");
-            if config_file.exists() {
-                if !quiet {
-                    println!("Loading startup commands from {config_file:?}");
-                }
-                let contents = std::fs::read_to_string(config_file)?;
-                outputs.merge(self.execute(&contents)?);
-            }
-            // Note: Loaded *after* init.evcxr so that it can access `:dep`s (or
-            // any other state changed by :commands) specified in the init file.
-            let prelude_file = config_dir.join("prelude.rs");
-            if prelude_file.exists() {
-                if !quiet {
-                    println!("Executing prelude from {prelude_file:?}");
-                }
-                let prelude = std::fs::read_to_string(prelude_file)?;
-                outputs.merge(self.execute(&prelude)?);
-            }
+        let init_config = InitConfig::parse_as_one_step()?;
+        if let Some(init) = init_config.init {
+            outputs.merge(self.execute(&init)?);
+        }
+        if let Some(prelude) = init_config.prelude {
+            outputs.merge(self.execute(&prelude)?);
         }
         Ok(outputs)
     }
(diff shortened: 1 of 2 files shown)


evcxr/src/eval_context.rs
```diff
@@ -102,6 +102,108 @@ pub(crate) struct Config {
...
+    pub(crate) prelude: Option<String>,
+}
+
+impl InitConfig {
+    pub(crate) fn check_if_exists(path: &Path) -> bool {
+        if !path.is_dir() {
+            return false
+        }
+        path
+            .read_dir()
```
davidlattimore: Rather than reading the directory to see if it contains the file, what about just doing `path.join("init.evcxr").exists()`?

evcxr/src/command_context.rs
```diff
@@ -225,27 +226,14 @@ Panic detected. Here's some useful information if you're filing a bug report.
         Ok(completions)
     }
 
-    fn load_config(&mut self, quiet: bool) -> Result<EvalOutputs, Error> {
+    fn load_config(&mut self, _quiet: bool) -> Result<EvalOutputs, Error> {
```
davidlattimore, line 229: Might as well remove the quiet argument if it's no longer used. Although I guess that possibly means that the --quiet flag now does nothing?

davidlattimore (commented):
Searching for a config file in the current directory seems like a great idea, thanks!

I'm a bit unsure about introducing a custom configuration file format though.

We could use toml. e.g. have an `evcxr.toml`. We already depend on the toml crate, so this wouldn't add any new dependencies. The toml file could configure the temporary directory. Putting `:dep` commands and Rust code in a config file probably doesn't make sense, but it'd be OK to have the config file specify a path to an `init.evcxr` file to load on startup. What do you think?

baiguoname:
> Searching for a config file in the current directory seems like a great idea, thanks!
> 
> I'm a bit unsure about introducing a custom configuration file format though.
> 
> We could use toml. e.g. have an `evcxr.toml`. We already depend on the toml crate, so this wouldn't add any new dependencies. The toml file could configure the temporary directory. Putting `:dep` commands and Rust code in a config file probably doesn't make sense, but it'd be OK to have the config file specify a path to an `init.evcxr` file to load on startup. What do you think?

Here is my understanding of what you said:
We should have a file named `evcxr.toml` located in the current directory. The contents should look like this:
```
[tmpdir]
path = "tmpdir"

[init]
path = "/root/arcta/init.evcxr"

[prelude]
path = "/root/arcta/prelude.rs"
```
If we don't have the file, then `evcxr` will execute the command `:load_config` as before.
If I'm understanding correctly, I thought this way had an advantage over above: I can have many notebooks in one directory, while each of them can have its own temporary directory, `init.evcxr` or `prelude.rs`. If you think this approach is fine, I can update this pr.

evcxr/src/eval_context.rs
baiguoname: It seems more reasonable and logical, thank you very much.

evcxr/src/command_context.rs
baiguoname: I think I should make it work the same as before, that is, to print the location of the files.

davidlattimore:
Yep, that's the kind of thing I had in mind, thanks for confirming. If we can't think of any further options that we might want for each of those sections, then we could potentially collapse them. e.g. something like:

```toml
[config]
tmpdir = "tmpdir"
init = "/root/arcta/init.evcxr"
prelude = "/root/arcta/prelude.evcxr"

But I don't feel strongly either way, so whatever you think is best.
```

davidlattimore:
Thanks for making those changes. I'll probably switch to using the `toml` crate to parse the config, but that can be done later.

Issue #353, opened 2023-11-26
Race condition between idle and iopub messages in `evcxr_jupyter`?


thomasjm:
Hello, I've been seeing the following intermittent issue. I'm not sure if it's due to a problem in `evcxr_jupyter` or not, but I'd love to hear what you think. I have code like the following:

```rust
use rand::prelude::*;
let x: u8 = random();
println!("{}", x);
```

I'm trying to package up `evcxr_jupyter` as part of a Jupyter kernel package, so I want to write a test that this code runs successfully. I run this code using [Papermill](https://github.com/nteract/papermill), and every once in a while (about 1 in 25 test runs) I fail to get any stdout.

My hypothesis is that the following is happening:
1) `evcxr` receives the execute command and starts running the code
2) (sometimes) `evcxr` sends the idle message before sending the stdout message
3) Papermill gets the idle message and decides to shut down the kernel. It sends the shutdown message and `evcxr_jupyter` shuts down, possibly never sending the stdout.

As far as I understand the Jupyter spec, kernels are supposed to make sure that all iopub messages are written to the socket before sending the kernel idle message. Is it possible that this is happening in a racy way in `evcxr_jupyter`?

davidlattimore:
Hi @thomasjm, thanks for the report. Yeah, that sounds about right. The iopub is handled by a separate thread, but when execution finishes, it acquires a lock on the iopub socket and sends the idle message (function `handle_shell_message` in core.rs).

Output from stdout and stderr are handled in `start_output_pass_through_thread`. It needs to be handled somewhat separately from execution because the user's code can continue to print stuff even it finishes execution - i.e. if it spawns a thread. Although this currently only works for stderr due to limitations in the way evcxr handles stdout (which should probably be fixed).

```rust
std::thread::spawn(|| {
    for i in 0..10 {
        eprintln!("{i}");
        std::thread::sleep(std::time::Duration::from_millis(500));
    }
})
```

One possible fix would be to change the channel that evcxr uses to send output to the IO thread to include events that mark the start and end of execution. The IO thread could then lock the iopub socket for the duration of execution. This would prevent `handle_shell_message` from sending its idle message on iopub until after the IO thread had sent the last bit of output from during execution.

thomasjm:
Thanks for explaining! FWIW, here's what the spec has to say about iopub from background threads:

[...]

PR #354, opened 2023-11-27
jupyter: Ensure iopub idle isn't sent before stdout output

davidlattimore:
Because we use stdout to notify completion of the user code, we can ensure that any output printed to stdout prior to the user code completing is sent to the iopub before we send the idle message.

We can't really do a similar thing with stderr because there's no control message sent to stderr to indicate the end of execution. However doing this at least for stdout is better than nothing.

Issue #353

### DIFF
--- a/evcxr/examples/example_eval.rs
+++ b/evcxr/examples/example_eval.rs
@@ -7,6 +7,7 @@
 
 use evcxr::Error;
 use evcxr::EvalContext;
+use evcxr::StdoutEvent;
 
 fn main() -> Result<(), Error> {
     // You must call ```evcxr::runtime_hook()``` at the top of main, otherwise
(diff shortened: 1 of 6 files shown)


thomasjm:
Hey, thanks again for fixing this issue. The solution seems to work, so I was just wondering if it'll end up in a release anytime soon?

> We can't really do a similar thing with stderr because there's no control message sent to stderr

Possible dumb question, but would it make sense to add an `eprintln!("{EVCXR_EXECUTION_COMPLETE}")` next to the existing `println!`?

https://github.com/evcxr/evcxr/blob/8fed0c29790963195647e7121cb621beda891653/evcxr/src/runtime.rs#L85

davidlattimore:
I've just done a release.

Regarding adding an execution-complete marker to stderr, I'm not sure. The use of markers in stdout currently causes various problems there, such as it ending up line-buffered, which means it's currently impossible to progressively update a line. Some of those issues might be fixed by flushing stdout straight away if we detect that the content already on the line can't be the start of a control instruction. It does make the code more complicated though. So I'm somewhat reticent to do the same thing to stderr.

thomasjm:
Thanks so much!

Hmm, I wonder if a better alternative would be to manage the concurrency without using the output streams for signalling at all. I'll poke around and see if something like that might be possible when I get a minute.

Issue #147, opened 2020-12-01
Rust, gpu and google colab


maurokenny:
Hi guys,

It is possible to use rust in google colab? In addition, is it possible to use also a gpu there with rust?

Thanks

davidlattimore:
Sorry about the slow reply. It is possible to use with colab. It's a super streamlined process though. You need a to build the evcxr_jupyter binary, then upload it to Google Drive and then reference the Google Drive file. See [this comment](https://github.com/google/evcxr/issues/112#issuecomment-609366627). I'd been meaning to try to figure out if the process could be streamlined somewhat. e.g. do automatic builds pushed to Google Drive.

As far as using the GPU, not sure sorry. I haven't done GPU programming from Rust or otherwise.

melMass:
> It is possible to use rust in google colab? In addition, is it possible to use also a gpu there with rust?

Yes and yes. Check this [gist](https://gist.github.com/korakot/ae95315ea6a3a3b33ee26203998a59a3), in the comments korakot provides a starting notebook (don't create one from scratch see comments for details)

gabelluardo:
Seems that the gist doesn't work anymore

davidlattimore:
From the limited reading I just did online, it seems that kernels other than Python 3 aren't supported by Colab. I found a link claiming to create a Swift Colab, but it too didn't seem to work. Perhaps it was possible before with some hacks but those hacks have stopped working with recent versions of Colab. I attempted to repeat those hacks just now, but if I try to execute Rust code, I get `Failed to execute cell. Could not send execute message to runtime: Error: await connected: disconnected`. Probably not much more that can be done at this point.

Panakotta00:
I've got it working again (kinda hacky tho), can be found in a comment of [this gist](https://gist.github.com/korakot/ae95315ea6a3a3b33ee26203998a59a3).

eatPizza311:
@wiseaidev Thank you for your amazing contribution.
The script works perfectly on CPU runtime, but when I tried on runtime with GPU it's not working.
Here is the error message:
![圖片](https://github.com/evcxr/evcxr/assets/43287234/1b4fb3c5-6751-496d-91bf-e5b5bca7f990)
After some digging, I found the symbolic link was correctly created, but the command is only working with the absolute path.
The configuration works perfectly if I prefixed the `nix-dev` and `evcxr_jupyter` with `/root/.nix-profile/bin` until line 26:
[...]

PR #359, opened 2023-12-20
add colab setup script

wiseaidev:
### Description
Related to #147 

This will help friends to execute rust code in colab.
image

### Demo

https://github.com/evcxr/evcxr/assets/62179149/479714ba-3f5c-41dc-8e30-b1a02ba6f92c

### Sample

[Open In Colab](https://colab.research.google.com/github/wiseaidev/rust-data-analysis/blob/main/7-calculus-tutorial-part-1.ipynb)

**Note**: You have to merge into main first to test these changes because of the path.

### DIFF
--- a/evcxr_jupyter/samples/evcxr_jupyter_tour.ipynb
+++ b/evcxr_jupyter/samples/evcxr_jupyter_tour.ipynb
@@ -4,6 +4,32 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
+    "# Google Colab Rust Setup\n",
+    "\n",
+    "The following cell is used to set up and spin up a Jupyter Notebook environment with a Rust kernel using Nix and IPC Proxy. "
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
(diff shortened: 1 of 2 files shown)


evcxr_jupyter/samples/evcxr_jupyter_tour.ipynb
```diff
@@ -1,5 +1,31 @@
...
+    "The following cell is used to set up and spin up a Jupyter Notebook environment with a Rust kernel using Nix and IPC Proxy. "
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "!wget -qO- https://gist.github.com/wiseaidev/2af6bef753d48565d11bcd478728c979/archive/3f6df40db09f3517ade41997b541b81f0976c12e.tar.gz | tar xvz --strip-components=1\n",
```
davidlattimore, line 18: Would it be possible to put the setup script in the evcxr repository? That would mean that if the file needs editing, we can just edit it via a regular github PR. Ideally the script could just live in the same directory as this file, although that would presumably only work if colab checks out the whole repository when it loads an ipynb file from github, but hopefully that's what it does.

wiseaidev: Got it. But, it turns out it only copies the ipynb file and not the whole directory:
Screenshot from 2023-12-20 23-50-50

The script lives in [this gist](https://gist.github.com/wiseaidev/2af6bef753d48565d11bcd478728c979) that I currently maintain. I will try my best to keep it up to date!

PR #381, opened 2024-08-15
Using crate `toml` to parse the startup configuration.

baiguoname:
This [discusstion ](https://github.com/evcxr/evcxr/discussions/368) mentioned this. 
In the previous method of parsing startup configuration, I have to create three files: `evcxr.toml`, `prelude.rs`, `init.evcxr`.  With this pr, only one file need to be created, named `evcxr.toml`, which may contain content like this: 
```
[dependencies]
rand = "*"
serde = { version = "1", features = [ "derive" ] } //this block can have multiple lines, the old version command :dep can not.

[evcxr]
tmpdir = "./tmpdir"
opt_level = "2"
prelude = """
use serde::{ Serialize, Deserialize };
use std::collections::HashMap as hm;
"""
```
The location of this file is the same as the old one `evcxr.toml`.
And the more important is that it's extensible


davidlattimore:
Thanks, the toml parsing seems nicer now!

I'd like to not break users' existing configurations though. So if a user has an `init.evcxr` or a `prelude.rs`, I'd like for that to keep working. At the moment, it seems that this removes support for reading either of these files. Would it be possible to add back reading of these files?

baiguoname:
> Thanks, the toml parsing seems nicer now!
> 
> I'd like to not break users' existing configurations though. So if a user has an `init.evcxr` or a `prelude.rs`, I'd like for that to keep working. At the moment, it seems that this removes support for reading either of these files. Would it be possible to add back reading of these files?

Ok，I'm going to add previous configurations method to this.

baiguoname:
I pushed a new commit, and I do give the precedence to `evcxr.toml` over `init.evcxr` and `prelude.rs`.

PR #395, opened 2025-02-16
feat: add shell commands support

drendog:
This PR duplicates #343, which looks to be unmaintained since 2023. Rather than waiting for the original PR to be updated, I cherry-picked its commits. Additionally, I fixed an issue where the output wasn't showing on jupyter notebook.

### DIFF
--- a/evcxr/src/command_context.rs
+++ b/evcxr/src/command_context.rs
@@ -9,6 +9,7 @@ use crate::code_block::CodeBlock;
 use crate::code_block::CodeKind;
 use crate::code_block::CommandCall;
 use crate::code_block::Segment;
+use crate::code_block::ShellCommand;
 use crate::code_block::{self};
 use crate::crash_guard::CrashGuard;
 use crate::errors::bail;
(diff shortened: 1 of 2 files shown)


evcxr/src/command_context.rs
```diff
@@ -185,13 +185,17 @@ Panic detected. Here's some useful information if you're filing a bug report.
...
-                // Add the stdout and stderr to eval_outputs
-                eval_outputs
-                    .content_by_mime_type
-                    .insert("stdout".to_string(), stdout_str.into());
+                let combined_output = if stderr_str.is_empty() {
+                    stdout_str.into_owned()
+                } else if stdout_str.is_empty() {
+                    stderr_str.clone().into_owned()
+                } else {
+                    format!("{}{}", stdout_str, stderr_str)
```
davidlattimore, line 193: It's probably simplest to just always do this - i.e. remove the above if and if-else.

davidlattimore (commented):
Thanks for looking at this!

Could you also have a look at addressing the other comment from the original PR? https://github.com/evcxr/evcxr/pull/343#discussion_r1390507917

drendog:
> Could you also have a look at addressing the other comment from the original PR? [#343 (comment)](https://github.com/evcxr/evcxr/pull/343#discussion_r1390507917)

image
I can't reproduce this bug, and I'm not able to get any errors or warnings on the code blocks with red squiggles. Am I missing something?

davidlattimore:
>  I can't reproduce this bug, and I'm not able to get any errors or warnings on the code blocks with red squiggles. Am I missing something?

I'm not sure from the screenshot, but it looks like you might be using Jupyter from vscode, which unfortunately doesn't support red squiggles. Try running `jupyter notebook` from a terminal, then opening it in a web browser.

drendog:
> > I can't reproduce this bug, and I'm not able to get any errors or warnings on the code blocks with red squiggles. Am I missing something?
> 
> I'm not sure from the screenshot, but it looks like you might be using Jupyter from vscode, which unfortunately doesn't support red squiggles. Try running `jupyter notebook` from a terminal, then opening it in a web browser.

I'm currently running jupyter notebook web frontend from browser, and the screenshot is from jupyter notebook (dark theme).

It looks like cargo check or something, is not being run in Jupyter code blocks, so I can't get any errors or warnings in any case

davidlattimore:
OK, I'll merge this and see if I can fix the unwanted red squiggle afterwards

wangjiawen2013:
Can we use cmd commands on windows OS ?

davidlattimore:
From a look at the changes in this PR, it's running `sh`, which I'm guessing probably isn't available on Windows, at least not by default. Although  - perhaps give it a try? Assuming it doesn't work, you can probably try changing the bit of code that invokes `sh` to do something appropriate for Windows and select which code to run with a `cfg`

PR #402, opened 2025-03-18
Revert "jupyter: Ensure iopub idle isn't sent before stdout output

### LINKED ISSUE #400: Cell won't stop on compilation error
I'm using evcxr_jupyter kernel in VSCode (both with local kernel and with existing Jupyter server). Cell won't stop executing on compilation error, restarting kernel does help, but the I have to compile dependencies again.

There is no such issue when using Colab. While using Jupyter notebook cells with compile errors won't stop either, however they dont interrupt other cell and they can be run without any problems. In VSCode other cells cant be executed.

The issue is present on my Mac and Windows (using WSL) machines

davidlattimore:
This reverts commit 88ebb9ee4e84ae53f1dab8064c864ec7420c4971.

Fixes #400

### DIFF
--- a/evcxr/examples/example_eval.rs
+++ b/evcxr/examples/example_eval.rs
@@ -7,7 +7,6 @@
 
 use evcxr::Error;
 use evcxr::EvalContext;
-use evcxr::StdoutEvent;
 
 fn main() -> Result<(), Error> {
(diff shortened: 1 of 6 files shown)


davidlattimore:
Went with #403 instead.

PR #414, opened 2025-07-19
Block until stdout_sender is empty with exponential backoff

thomasjm:
This is the fix discussed in #353.

As I mentioned there, I tested this with 100 test runs and it seems to fix the issue. But, I'm not familiar enough with the workings here to be sure this will work in all circumstances. For example, a) background threads continuing to write to stdout and b) possible concurrent use of `stdout_sender` between different Jupyter requests.

### DIFF
--- a/evcxr/src/eval_context.rs
+++ b/evcxr/src/eval_context.rs
@@ -950,6 +950,14 @@ impl EvalContext {
         } else if !lost_variables.is_empty() {
             return Err(Error::TypeRedefinedVariablesLost(lost_variables));
         }
+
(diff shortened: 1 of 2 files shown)


davidlattimore:
Thanks for the PR! It looks like the tests timed out. The test that timed out was called `printing` in `integration_tests.rs`. I'd say the easiest fix is likely to change the test to spawn a thread or two - one to do the eval and one to read the outputs, then join those two threads. Let me know if you need help making that change.

thomasjm:
Thanks @davidlattimore! If you're happy with how this works, would you mind cutting a release with it?

PR #421, opened 2025-08-28
Use .display() for :last_compile_dir

kryptan:
`:last_compile_dir` currently uses `Debug` implementation for `Path` which replaces each backslash `\` with `\\`  for Windows paths. This makes it inconvenient as one needs to replace them back to navigate to such path. Instead we can use [`Path::display`](https://doc.rust-lang.org/std/path/struct.Path.html#method.display) method to properly display paths. VSCode now turns it into a hyperlink and offers to open it in VSCode if you click on it which is very useful.

Before:
before

After:
after


davidlattimore:
Thanks

PR #449, opened 2026-05-06
Add --config <PATH> flag to override init file path

### LINKED ISSUE #393: (feature request) adding optional config location to cli
i was wondering if it would be possible to specify a path to a config file instead of the default one? I think i'll take a shot at it myself, but in general would that be logical for the project?

CGMossa:
## Summary

- Adds `--config <PATH>` CLI flag to the `evcxr` REPL binary
- When provided, loads that file as the init script instead of the default `~/.config/evcxr/init.evcxr`
- Errors out with a clear message and non-zero exit if the given path does not exist; default behavior (no flag) is unchanged

Resolves #393.

## Test plan

- [ ] `evcxr --config /path/to/my-init.evcxr` — should load commands from that file on startup and print `Startup commands will be loaded from /path/to/my-init.evcxr`
- [ ] `evcxr --config /nonexistent/path` — should print `--config: file not found: /nonexistent/path` and exit nonzero immediately (before rustyline starts)
- [ ] `evcxr` (no flag) — default behavior unchanged; still loads `~/.config/evcxr/init.evcxr` if it exists, or silently skips if absent

### DIFF
--- a/evcxr_repl/src/bin/evcxr.rs
+++ b/evcxr_repl/src/bin/evcxr.rs
@@ -258,8 +258,8 @@ struct Options {
     #[clap(long, default_value = "emacs")]
     edit_mode: EditMode,
 
-    /// Path to an init file to load instead of the default ~/.config/evcxr/init.evcxr.
-    /// Errors if the file does not exist.
+    /// Path to an init file to load instead of the platform default. Errors if the file does
+    /// not exist.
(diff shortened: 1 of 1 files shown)


CGMossa:
Closing — moving iteration to fork while we validate. See CGMossa/evcxr#2. Apologies for the noise.

Issue #376, opened 2024-07-17
Add ability to enable feature when running


leighmcculloch:
When writing code that uses features, or copy-pasting in code that uses features, or using macros that wrap code in different features, it would be helpful if the notebook also offered an option to enable which features are enabled for the current crate.

davidlattimore:
Do you mean enable features of the crates that you're using? That can be done like this:

```
:dep regex = { version = "1.10.5", default-features = false, features = ["std"] }
```

leighmcculloch:
I was meaning for the crate that the notebook is creating from the code in the panel.

The code I'm typing into the panel has `cfg(features = "...")` portions in it. I can manually comment / uncomment the code, but I thought there might be a way to list an enabled feature with an option.

In another snippet I have a macro is creating features, and I'd like to enable that feature.

davidlattimore:
Ah, gotcha. You're correct that there's not currently a way to do that. Seems like a reasonable if slightly unusual thing to do. If you'd like to have a try at implementing it, I'd be happy to provide guidance as needed

leighmcculloch:
It is a bit unusual, which also makes me think if I was maintaining this runtime that it wouldn't be ideal to add for a rarely use feature. Even though I'm trying to do it now, I can't imagine how many folks actually need to do this. If you think it makes sense for the future direction I'll take a stab at it.

davidlattimore:
Although it's unusual, I don't think it should add any substantial complexity or change the way things work, so I think it's fine to add.

It looks like having `cfg!(feature = "foo")` with no feature `foo` declared in `Cargo.toml` doesn't issue any errors. So it should be possible to declare and enable features just by making changes to the `Cargo.toml` that gets generated. e.g.

```toml
[features]
default = ["foo"]
foo = []
```

Any features that aren't enabled can just be ignored.

You could have separate commands for enabling and disabling features, but it's probably simplest just to have one command that sets which features are enabled. e.g. `:features foo,bar`. You can then plumb that list of features through to `ContextState`. You can then change the code that generated `Cargo.toml` which is `get_cargo_toml_contents` in `module.rs`.

PR #450, opened 2026-05-06
Add :features command for notebook crate (#376)

CGMossa:
## Summary

- Adds a `:features` command that lets users declare features on the generated evaluation crate, so `cfg(feature = "foo")` gating works in notebook code.
- Plumbs a `Vec<String>` through `Config` / `ContextState` into `get_cargo_toml_contents` in `module.rs`, emitting a `[features]` section with all declared features listed in `default` (so they activate without any explicit `--features` flag).
- The `[features]` section is omitted entirely when the list is empty — no behaviour change for existing users.

Implementation follows the sketch in #376:
> declare features in the generated Cargo.toml … plumb a Vec<String> through ContextState, modify get_cargo_toml_contents … add a :features foo,bar command

## Usage

```
:features foo,bar     # set features (replaces previous list)
:features             # print currently enabled features
:features             # (with trailing space / empty arg) clears all features
```

## Test plan

- [ ] Start the REPL or a Jupyter kernel.
- [ ] Run `:features myfeat` then evaluate `cfg!(feature = "myfeat")` — should print `true`.
- [ ] Run `:features` (no args) — should print `features: myfeat`.
- [ ] Run `:features ` (empty arg) — should print `features: (none)`.
- [ ] Evaluate `cfg!(feature = "myfeat")` again — should now print `false`.
- [ ] Verify that a session that never calls `:features` produces the same Cargo.toml as before (no `[features]` section).

### DIFF
--- a/evcxr/src/command_context.rs
+++ b/evcxr/src/command_context.rs
@@ -537,6 +537,34 @@ Panic detected. Here's some useful information if you're filing a bug report.
                 },
             ),
             AvailableCommand::new(
+                ":features",
+                "Set/clear features for the generated crate. e.g. :features foo,bar",
+                |_ctx, state, args| {
+                    match args.as_deref() {
+                        // No argument: print current features.
+                        None => {
+                            let current = state.features().join(", ");
+                            text_output(format!("features: {current}"))
+                        }
+                        // Empty string (":features " with trailing space): clear features.
+                        Some("") => {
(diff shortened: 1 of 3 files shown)


CGMossa:
Closing — moving iteration to fork while we validate. See CGMossa/evcxr#3. Apologies for the noise.

PR #454, opened 2026-05-06
Add --script flag for running a file (#188)

### LINKED ISSUE #188: Script file support
I'd like to be able to pass in a script file to evcxr for execution; something like

```
evcxr script.rs
```

(though maybe script files should have a different extension)

Such functionality should also optionally allow running commands interactively after the script executes. This would allow the use of script files as a sort of prelude for setting up the interactive environment. Currently, the only way I know of to set up the interactive environment in a predefined way is through copy-pasting commands.

CGMossa:
## Summary

- Add `--script <FILE>` flag to `evcxr_repl` that reads the given file and executes each line through `CommandContext` before entering the interactive REPL, mirroring how a user would type them at the prompt.
- Add `--exit-after-script` flag: when set, the binary exits cleanly after the script finishes instead of opening the readline loop.
- File-not-found produces a hard error with a clear message; errors during script execution use the existing `repl.execute` path (print and continue).

Closes #188.

## Test plan

- [ ] Create a test script `hello.evcxr` containing `println!("hello from script");`
- [ ] Run `evcxr --script hello.evcxr` and verify the output is printed and the REPL starts.
- [ ] Run `evcxr --script hello.evcxr --exit-after-script` and verify output is printed and the binary exits.
- [ ] Run `evcxr --script nonexistent.evcxr` and verify a clear error message is shown.
- [ ] Run `evcxr` (no flags) and verify the REPL still works as before.

### DIFF
--- a/evcxr_repl/src/bin/evcxr.rs
+++ b/evcxr_repl/src/bin/evcxr.rs
@@ -250,6 +250,14 @@ struct Options {
     #[clap(long, default_value = "emacs")]
     edit_mode: EditMode,
 
+    /// Script file to execute before entering the interactive REPL.
+    #[clap(long)]
+    script: Option<std::path::PathBuf>,
+
+    /// Exit after running the script instead of dropping into the interactive REPL.
+    #[clap(long)]
+    exit_after_script: bool,
+
     /// Extra arguments; ignored, but show up in std::env::args() which is passed to the subprocess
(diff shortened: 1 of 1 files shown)


CGMossa:
Closing — moving iteration to fork while we validate. See CGMossa/evcxr#7. Apologies for the noise.

PR #455, opened 2026-05-06
Show compiler warnings in REPL (#238)

### LINKED ISSUE #238: Show compiler warnings
Evcxr doesn't currently show them if the compilation is successful, and I didn't find any option for enabling warnings, and any issue about that (but sorry if there is and it is a duplicate).

Warnings are useful, because people can make mistakes in jupyter notebooks as well, and in my usecase (I'm writing a rust tutorial) I want to intentionally trigger a warning and show it to the reader. (I currently workaround it by `#[deny]` the warning)

Is there a fundamental limit preventing that? I think an option for enabling them is fine too if they are considered unnecessary noise by default.

CGMossa:
## Summary

- Rustc warnings from user code are now shown in the REPL after a successful compile, matching the behaviour already present in the Jupyter kernel.
- On a successful build, `warnings_from_cargo_output` parses warning-level JSON messages (user-code only) from the cargo output and attaches them to the new `EvalOutputs.warnings` field.
- The REPL frontend prints each warning using rustc's pre-formatted `rendered` string, which carries the correct yellow `warning:` prefix and source context.

Fixes #238. Jupyter behaviour is unchanged.

## Test plan

- [ ] Start the REPL and run `fn foo() { let mut s = String::new(); }` — expect an `unused variable` warning to be printed immediately after the successful compilation.
- [ ] Run `let _x = 1;` — expect no warning (underscore-prefixed variables suppress warnings).
- [ ] Run code that has a compile error — expect unchanged error display, no regressions.
- [ ] Run code with no warnings — expect clean output as before.

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/evcxr/src/eval_context.rs
+++ b/evcxr/src/eval_context.rs
@@ -806,7 +806,8 @@ impl EvalContext {
         }
         phases.phase_complete("Final compile");
 
-        let output = self.run_and_capture_output(state, &so_file, callbacks)?;
+        let mut output = self.run_and_capture_output(state, &so_file, callbacks)?;
+        output.warnings = so_file.warnings;
         Ok(ExecutionArtifacts { output })
     }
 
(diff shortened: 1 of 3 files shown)


CGMossa:
Closing — moving iteration to fork while we validate. See CGMossa/evcxr#8. Apologies for the noise.

PR #468, opened 2026-05-20
Feat/lsp-rich-completions

### LINKED ISSUE #466: Enhance rust-analyzer support
I'm moving a lot of my Python work over to Rust and would like some better rust-analyzer integration. If we look at the situation with Python, we get -

- doc strings
- type info 
- function signatures, or init signatures for class init methods

Rust should be even better at this given the stronger typing, but the experience I'm having with completions and documentation in the notebook is pretty poor.

This may be a dupe of [this](https://github.com/evcxr/evcxr/issues/102) issue, but it is very old and I think a lot of the discussion is out of date.

I'm happy to work on this (it'll be AI assisted as I'm not that familiar with this code base) if it is of interest to the community.

Atreus-Technologies:
Richer completions for LSP requests. Now including 
- Types
- Docstrings where available

Previously, shift+tab didn't work the way we'd expect, now it does.

Manually tested, looks good from there. 

Feedback welcome, I'm not amazing at rust, and not familiar with this repo. But I've taken a look at the tests and they seem valid.

Totally AI but I've read for sanity and I think it looks useful. I think a squash and merge would be appropriate here but can squash on my end if that's preferred, let me know.

Will close #466.

### DIFF
--- a/evcxr/tests/integration_tests.rs
+++ b/evcxr/tests/integration_tests.rs
@@ -1180,21 +1180,11 @@ fn completion_fields_populated() {
         );
     }
 
-    // The `reserve` completion should carry a type signature in detail.
     let reserve = completions
         .completions
         .iter()
         .find(|c| c.code == "reserve(additional)")
         .expect("expected reserve(additional) in completions");
-    assert!(
-        reserve.detail.is_some(),
-        "reserve should have a detail (type signature)"
-    );
-    let detail = reserve.detail.as_ref().unwrap();
-    assert!(
-        detail.contains("usize"),
-        "reserve detail should mention its argument type, got: {detail}"
-    );
     assert_eq!(
         reserve.kind, "function",
         "method should have kind 'function'"
@@ -1216,14 +1206,6 @@ fn command_completion_fields() {
     assert_eq!(
         dep.kind, "magic",
         "command completions should have kind 'magic'"
-    );
-    assert!(
-        dep.detail.is_none(),
-        "command completions should have no detail"
-    );
-    assert!(
-        dep.documentation.is_none(),
-        "command completions should have no documentation"
     );
 }
 
--- a/evcxr/tests/integration_tests.rs
+++ b/evcxr/tests/integration_tests.rs
@@ -1195,6 +1195,10 @@ fn completion_fields_populated() {
         detail.contains("usize"),
         "reserve detail should mention its argument type, got: {detail}"
     );
+    assert_eq!(
+        reserve.kind, "function",
+        "method should have kind 'function'"
+    );
 }
 
 #[test]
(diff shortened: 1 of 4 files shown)


evcxr/tests/integration_tests.rs
```diff
@@ -892,7 +892,6 @@ fn simple_completions(ctx: &mut CommandContext, code: &str) -> HashSet<String> {
 }
 
 // Not sure why this is failing on mac
-#[cfg(not(target_os = "macos"))]
```
davidlattimore, line 895: If you're removing this exclusion, then we should remove the comment above it as well. Although removing this isn't really related to the rest of the PR, I don't object to removing it in this PR given that it appears the test passed OK on mac. Possibly recent updates to rust-analyzer have allowed the test to pass.

evcxr_jupyter/src/core.rs
```diff
@@ -761,22 +773,63 @@ async fn handle_completion_request(
...
-            .collect();
+        let cursor_start = byte_offset_to_grapheme_offset(code, completions.start_offset)?;
+        let cursor_end = byte_offset_to_grapheme_offset(code, completions.end_offset)?;
+        let mut matches: Vec<String> = Vec::new();
+        let mut type_metadata: Vec<JsonValue> = Vec::new();
+        for c in &completions.completions {
+            matches.push(c.code.clone());
+            type_metadata.push(object! {
+                "text"      => c.label.clone(),
+                "type"      => "function",
```
davidlattimore: Should this always be "function"?

davidlattimore (commented):
Thanks. I've tried it out and it works for me.

Looks like the only CI failure was for formatting, so you probably just need to run `cargo fmt` or similar.

evcxr_jupyter/src/core.rs
Atreus-Technologies: Good catch, that's definitely wrong. Fixing it now, apologies.

evcxr/tests/integration_tests.rs
Atreus-Technologies: Somewhat unrelated, although with the most recent rust stable it seems that both of the tests that were excluded on mac previously now pass. I've removed both exclusions but if you'd rather I reinstate them to keep the PR more on-scope then I can do that too.

davidlattimore: Removing it in this PR is fine by me so long as the out-of-date comment is removed too

Atreus-Technologies:
OK, I've made a few changes here. Especially simplifying, tightening up some tests, fixing the kind issue so that we see if something is a function or a "class" (this is the only language Jupyter has, so we'll have to deal with the slightly awkward match to Rust's language.)

I think it is a bit tighter now.

NB: let me do a squash obviously prior to merging, we don't want all these commits in the final merge.

davidlattimore (approved):
Looks good. Thanks. I can also just squash when merging if you like? Unless you had a particular way you wanted to do it? I often squash when merging anyway, since it means that the PR number gets included in the commit title, which is nice.

Atreus-Technologies:
> Looks good. Thanks. I can also just squash when merging if you like? Unless you had a particular way you wanted to do it? I often squash when merging anyway, since it means that the PR number gets included in the commit title, which is nice.

Works for me! The commits here are somewhat informative anyway so they can stick around in the PR after merge.

PR #502, opened 2026-08-23
Preserve invalid code for compiler diagnostics

### LINKED ISSUE #495: Some sort of weird parser bug when `in` appears in a function parameter name position
today I wrote `fn doit(in: u8) {}` into my favourite Rust REPL, absent-mindedly forgetting that `in` is a keyword. And I was hit back with the following diagnostic:
```rs
Error: unexpected closing delimiter: `}`
   ╭─[ command:1:1 ]
   │
 1 │ fn do_it(in: u8) {}
   │                │ 
   │                ╰─ missing open `(` for this delimiter
───╯
```
_...wat._

So, utterly bewlidered, I very nearly submitted a bug report to rust-lang/rust, before double checking and finding rustc gives a perfectly reasonable error response for that code:
```rs
error: expected parameter name, found keyword `in`
 --> src/main.rs:1:10
  |
1 | fn do_it(in: u8) {}
  |          ^^ expected parameter name
```
So, investigating a bit more, I opened up my `:last_compile_dir`, to have a look at the code Evcxr generated:
```rs
#![allow(unused_imports, unused_mut, dead_code)]
fn do_it(
#[unsafe(no_mangle)]
pub extern "C" fn run_user_code_3(
evcxr_variable_store: *mut u8) -> *mut u8 {
in
:
u8
) 
(
{}
).evcxr_display();
evcxr_variable_store
}
```
So, it appears the evcxr parser for detecting function definitions has somehow failed spectacularly. Let me know if there's any other info  I could provide

pttydou:
Fixes #495.

When rust-analyzer recovered from a parse error, Evcxr treated each recovery node as a separate statement. This change keeps the affected node and the remaining input together so rustc receives the original invalid function.

Tests:
- `cargo build -j 1`
- `cargo test -j 1 -- --test-threads 1`
- `cargo fmt --all -- --check`
- `cargo clippy -j 1`

### DIFF
--- a/evcxr/src/statement_splitter.rs
+++ b/evcxr/src/statement_splitter.rs
@@ -26,6 +26,11 @@ pub(crate) fn split_into_statements(code: &str) -> Vec<OriginalUserCode<'_>> {
         &(prelude.to_owned() + code + "\n}"),
         crate::rust_analyzer::EDITION,
     );
+    let first_error_byte = parsed_file
+        .errors()
+        .iter()
+        .map(|error| usize::from(error.range().start()).saturating_sub(prelude.len()))
+        .min();
     let mut start_byte = 0;
     if let Some(stmt_list) = parsed_file
         .syntax_node()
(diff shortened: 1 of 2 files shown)


davidlattimore (approved):
Thanks :)

PR #507, opened 2026-09-07
Fix comment handling when looking for commands

MaxFreedomPollard:
Comment lines are only skipped while evcxr looks for commands if they start at column zero, so an indented comment before a command ends the search and the command is then treated as Rust code. `ctx.check("    // An indented comment\n    :an_invalid_command")` reports a syntax error at the colon instead of "Unrecognised command", and the same input through `execute` fails to compile rather than running the command.

The check in `evcxr/src/code_block.rs` is also too broad in the other direction. `line.starts_with("//")` matches `///` too, so a doc comment at column zero is skipped and never reaches the compiler, and the item it documents ends up with no documentation. `check_for_doc` only passes because its `///this is my struct` happens to be indented; put it at column zero and `:doc MyStruct` prints the struct with the doc text gone.

`from_original_user_code` now trims leading whitespace before it looks for `//`, and an outer doc comment no longer ends the search on its own. It is held as the point where the code starts if code follows it, and dropped if a command follows, which is what happened before. `//!` and `////` stay plain comments: an inner doc comment would document the module evcxr generates around the cell rather than any of the user's code, and `////` is not a doc comment to rustc either.

Tests:
- `cargo test -p evcxr -- --test-threads 1`, on macOS with rustc 1.98.0: 32 unit tests and 48 integration tests pass.
- The four new unit tests and the two new integration tests fail on main. `indented_comment_before_command` finds no commands, `doc_comment_stays_with_its_item` loses the `///` line, `command_after_indented_comment` gets `error 2:5-2:6` instead of `error 2:1-2:20`, and `check_for_doc_starting_at_column_zero` gets `ctx\n\nstruct MyStruct(usize)` with the doc text missing. `doc_comment_before_command_is_dropped` and `comments_that_arent_outer_doc_comments_are_dropped` pin behaviour that has not changed and pass either way.
- `cargo fmt --all -- --check`
- `cargo clippy -p evcxr --all-targets`

No new dependencies, so `Cargo.lock` and `cackle.toml` are untouched.

### DIFF
--- a/evcxr/src/code_block.rs
+++ b/evcxr/src/code_block.rs
@@ -106,6 +106,15 @@ fn num_lines(code: &str) -> usize {
     code.chars().filter(|ch| *ch == '\n').count()
 }
 
+/// Returns whether `line`, ignoring any leading whitespace, starts an outer doc comment. Note that
+/// `////` starts a plain comment, not a doc comment. Inner doc comments (`//!`) aren't included,
+/// since they'd document the module that we generate around the user's code rather than anything
+/// the user wrote.
+fn is_outer_doc_comment(line: &str) -> bool {
+    let line = line.trim_start();
+    line.starts_with("///") && !line.starts_with("////")
+}
+
 pub(crate) fn count_columns(code: &str) -> usize {
     // We use characters here, not graphemes because seems to be how columns are counted by the rust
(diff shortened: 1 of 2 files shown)


davidlattimore:
Thanks :)

